PR #9, opened 2020-12-07
SQL Database

adzialocha:
* Agnostic SQL database backend via `sqlx` for PostgreSQL, MySQL and SQLite
* Introduce `DATABASE_URL` and `DATABASE_MAX_CONNECTIONS` environment variables
* Manage database via connection pool
* Close connections on shutdown
* Create / use SQLite database in data directory by default
* Embedded database migrations when running application


db/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+        .max_connections(max_connections)
+        .connect(&url)
+        .await?;
+
+    Ok(pool)
+}
+
+/// Run any pending database migrations from inside the application.
+pub async fn run_pending_migrations(pool: &Pool) -> Result<()> {
+    migrate!("./migrations").run(pool).await?;
```
cafca, line 33: Don't we need the `migrate` cargo feature flag for this?

https://github.com/launchbadge/sqlx#cargo-feature-flags

cafca: Could also be interesting for migration management: https://crates.io/crates/sqlx-cli

cafca: Also: where is this "./migrations" folder?

cafca (changes requested):
Just a couple of questions - but seriously, please no odd number for the default port!

README.md
```diff
@@ -18,6 +18,8 @@ OPTIONS:
 
 ## Environment variables
 
+* `DATABASE_URL` Database url (SQLite, MySQL, PostgreSQL) (default `sqlite:<data-dir>/main.db`).
+* `DATABASE_MAX_CONNECTIONS` Maximum number of database connections in pool (default `32`).
 * `HTTP_PORT` RPC API HTTP server port (default `9123`).
```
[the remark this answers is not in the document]
adzialocha: I haven't thought about the ports yet, we could have something fun maybe? I remember `2020` from a meeting.

One question: What do odd / even ports stand for?

db/src/utils.rs
adzialocha: > Don't we need the `migrate` cargo feature flag for this?
> 
> https://github.com/launchbadge/sqlx#cargo-feature-flags

It is set by default 🙂

adzialocha: > Also: where is this "./migrations" folder?

Its in the `db` folder but not checked in as it's empty, will add a `.gitkeep` 👍

README.md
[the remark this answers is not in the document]
cafca: odd=ugly even=precious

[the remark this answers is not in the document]
adzialocha: Also, I found this interesting bit on [Wikipedia](https://en.wikipedia.org/wiki/Network_socket#Definition):

> In the original definition of socket given in RFC 147, as it was related to the ARPA network in 1971, "the socket is specified as a 32 bit number with even sockets identifying receiving sockets and odd sockets identifying sending sockets." Today, however, socket communications are bidirectional.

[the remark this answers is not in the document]
adzialocha: Changed it for now to `2020` (HTTP) and `2022` (WebSocket). Funny to see what is registrated for: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=36

README.md
```diff
@@ -18,13 +18,13 @@ OPTIONS:
 
 ## Environment variables
 
-* `DATABASE_URL` Database url (SQLite, MySQL, PostgreSQL) (default `sqlite:<data-dir>/main.db`).
+* `DATABASE_URL` Database url (SQLite, MySQL, PostgreSQL) (default `sqlite:<data-dir>/p2panda-node.sqlite3`).
 * `DATABASE_MAX_CONNECTIONS` Maximum number of database connections in pool (default `32`).
-* `HTTP_PORT` RPC API HTTP server port (default `9123`).
+* `HTTP_PORT` RPC API HTTP server port (default `2020`).
```
cafca, line 23: ❣️

cafca (approved):

PR #11, opened 2020-12-14
Implement getEntryArguments RPC method

cafca:
* Add SQL migrations:
  * Introduce `logs` table holding state of which `log_id` is used for which `schema`
  * Introduce `entries` table storing all bamboo entries
* Add `models` module implementing sql methods for `Log` and `Entry`:
  * Implement method to find next free user log id for author
  * Implement method to retrieve bamboo entry at head or sequence number
* Add `types` module
  * Add basic data types like `LogId`, `Author` and `Schema`
  * Add validations for all types via `validator` crate
  * e2e typed data from RPC API to database
* Introduced ApiService backend and mpsc channel for communication with frontend
  * Implemented `getEntryArguments` (former `nextEntryArguments`) RPC call returning back and skiplink, schema's log id and last sequence number
* Add custom `Error` type via `thiserror` crate for correct handling of different database, RPC and validation errors, respond with nicely looking errors to RPC clients
* Added a bunch of tests and documentation
* Added `test_helpers` module for common testing methods, like creating an in-memory sqlite testing database


adzialocha:
Some more random thoughts and open questions:

* The `Entry` model is not 100% thought through (especially how to deal with the entry bytes), but I think this all will resolve when implementing `panda_publishEntry`.
* How do we encode the `Author`? Currently its hexadecimal (resulting in 64 chars), other options are Base64?
* How do we generate entry hashes?
* Should we move the `types` module into `p2panda-rs`?

cafca:
should add validation to prevent rpc calls with invalid or already used log ids

adzialocha:
> should add validation to prevent rpc calls with invalid or already used log ids

Realized that we don't need to check for the log id right now as the user can't set it themselves in the `getEntryArguments` call. Think adding the `assert` in the method itself was good, but no need to add any further user data validation.

node/src/errors.rs
```diff
@@ -0,0 +1,82 @@
...
+                        format!("`{}`: {}", error_kind.0, error_message)
+                    })
+                    .collect::<Vec<String>>()
+                    .join(", ");
+
+                jsonrpc_core::Error {
+                    code: jsonrpc_core::ErrorCode::InvalidParams,
+                    message: format!("Invalid params: {}.", message),
+                    data: None,
+                }
```
cafca, line 56: I need to remember this for when someone asks why rust is crazy ;)

node/src/lib.rs
```diff
@@ -8,14 +8,19 @@
     unsafe_code,
     unstable_features,
     unused_import_braces,
-    unused_qualifications,
+    unused_qualifications
```
cafca, line 11: why not?

adzialocha: why not the comma? :-)

adzialocha: I actually don't know! The auto formatter removes it for me ..

cafca: yes the comma - i expected trailing commas for multi-line enumerations. is this something we need to configure for rust specifically?

node/src/types/entry_hash.rs
```diff
@@ -0,0 +1,122 @@
...
+    #[test]
+    fn validate() {
+        assert!(EntryHash::new("abcdefg").is_err());
+        assert!(EntryHash::new("112233445566ff").is_err());
+        assert!(EntryHash::new(
+            "01234567812345678123456781234567812345678123456781234567812345678"
+        )
+        .is_err());
+        assert!(
+            EntryHash::new("004069db5208a271c53de8a1b6220e6a4d7fcccd89e6c0c7e75c833e34dc68d932624f2ccf27513f42fb7d0e4390a99b225bad41ba14a6297537246dbe4e6ce150e8").is_ok()
```
cafca, line 114: 😝

node/src/lib.rs
adzialocha: Ja, in this case its compiler flags, telling `rustc` to change its internal linters, maybe there its different?

PR #23, opened 2021-04-09
Add panda_queryEntries rpc method

cafca:
This is very experimental of course

- provides an rpc method `panda_queryEntries` that returns all entry db records
- the return value follows the database layout quite closely to allow for experimentation in the frontend client while having access to all available db fields. Do we want to keep such an rpc method? Should entries be decoded before returning them? 

# Todo
- [x] Find a meaningful way to restrict querying all entries
~~Add test with a db fixture so that the test actually returns entries~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,9 +8,9 @@ checksum = "739f4a8db6605981345c5654f3a85b056ce52f37a39d34da03f25bf2151ea16e"
 
 [[package]]
 name = "ahash"
-version = "0.6.3"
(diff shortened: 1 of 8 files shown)


adzialocha:
How cool is this? :sunflower: 

@cafca Regarding the schema: You could look up in the `Logs` table which logids + authors exist for that schema and then select the `Entries` accordingly.

cafca:
Rightttt. And here the sql fun begins!

PR #32, opened 2021-05-18
Fix for breaking changes in p2panda-rs

sandreae:
Fixes for when we merge this: https://github.com/p2panda/p2panda/pull/62

- use new encoder module for signing and decoding entries
- SeqNum is no longer optional in Entry
- SeqNum is now current sequence number not previous in Entry
- Message in no longer optional in Entry

### DIFF
--- a/aquadoggo/Cargo.toml
+++ b/aquadoggo/Cargo.toml
@@ -42,7 +42,7 @@ tag = "v0.1.0-pre-31"
 # crate via git. See: https://github.com/p2panda/p2panda/issues/38
 [dependencies.p2panda-rs]
 git = "https://github.com/p2panda/p2panda"
-branch = "main"
+branch = "sam-refactor-entry"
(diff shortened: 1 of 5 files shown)


aquadoggo/Cargo.toml
```diff
@@ -42,7 +42,7 @@ tag = "v0.1.0-pre-31"
 # crate via git. See: https://github.com/p2panda/p2panda/issues/38
 [dependencies.p2panda-rs]
 git = "https://github.com/p2panda/p2panda"
-branch = "main"
+branch = "sam-refactor-entry"
```
sandreae, line 45: This will need to be changed back to `main` once we merge everything.

sandreae:
Actually, this isn't ready yet... ;-p

sandreae:
Now it's ready I think.

adzialocha (approved):

PR #34, opened 2021-05-23
Update p2panda-rs imports

sandreae:
- update `p2panda_rs` imports
- change `as_hex()` to new `as_str()` method


adzialocha (approved):
Thank you! 🤩

PR #47, opened 2021-12-13
Make JSON RPC methods compatible with new document logs flow

adzialocha:
This PR prepares the aquadoggo to allow clients to work with the new p2panda specification, for example our new organization of logs *per document* and the new way of how we organize log ids.

* Remove distinction of system and application log ids
* Changes `nextEntryArgs` to allow an optional `document` param, this fixes a problem where we could not create new documents because of an chicken-egg situation (I don't know the document hash before I created it)
* Refactors *document per log* flow to resemble the following logic:
  ```
  Document x:
  log 1: [x] <- [y] <- [z] <- [l]              (key A)
  log 2:           \-- [i] <- [j] <- [k]       (key B)
  ```
  Note that this is not fully complete yet since we miss the `previousOperation` field in Operations. See related issue here: https://github.com/p2panda/aquadoggo/issues/49. For now it only works in single-writer scenarios.
* Removes the need of an `id` aka *documentId* aka *documentHash* inside the Operation, we can get it from the database instead

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/db/models/log.rs
+++ b/aquadoggo/src/db/models/log.rs
@@ -178,10 +178,8 @@ mod tests {
     use p2panda_rs::identity::{Author, KeyPair};
     use p2panda_rs::operation::{Operation, OperationEncoded, OperationFields, OperationValue};
 
-    use crate::{
-        db::models::Entry as dbEntry,
-        test_helpers::{initialize_db, random_entry_hash},
-    };
+    use crate::db::models::Entry as dbEntry;
+    use crate::test_helpers::{initialize_db, random_entry_hash};
 
     use super::Log;
 
@@ -222,10 +220,10 @@ mod tests {
     }
 
     #[async_std::test]
-    async fn instance_log_id() {
+    async fn document_log_id() {
         let pool = initialize_db().await;
 
-        // Create an instance
+        // Create a new document
         // TODO: use p2panda-rs test utils once available
         let key_pair = KeyPair::new();
         let author = Author::try_from(key_pair.public_key().clone()).unwrap();
@@ -241,16 +239,15 @@ mod tests {
         let entry = Entry::new(&log_id, Some(&operation), None, None, &seq_num).unwrap();
         let entry_encoded = sign_and_encode(&entry, &key_pair).unwrap();
 
-        // Expect no log id when instance not in database
-        // @TODO
-        /* assert_eq!(
-            Log::get_log_id_by_entry(&pool, &entry_encoded.hash())
+        // Expect database to return nothing yet
+        assert_eq!(
+            Log::get_document_by_entry(&pool, &entry_encoded.hash())
                 .await
                 .unwrap(),
             None
-        ); */
+        );
 
-        // Store instance in db
+        // Store entry in database
         assert!(dbEntry::insert(
             &pool,
             &author,
@@ -264,14 +261,21 @@ mod tests {
         .await
         .is_ok());
 
-        // Expect to find a log id for the instance
-        // @TODO
-        /* assert_eq!(
-            Log::get_log_id_by_entry(&pool, &entry_encoded.hash())
+        // Store log in database
+        assert!(
+            Log::insert(&pool, &author, &entry_encoded.hash(), &schema, &log_id)
+                .await
+                .is_ok()
+        );
+
+        // Expect to find document in database. The document hash should be the same as the hash of
+        // the entry which referred to the `CREATE` operation.
+        assert_eq!(
+            Log::get_document_by_entry(&pool, &entry_encoded.hash())
                 .await
                 .unwrap(),
-            Some(log_id)
-        ); */
+            Some(entry_encoded.hash())
+        );
     }
 
     #[async_std::test]
--- a/aquadoggo/src/db/models/log.rs
+++ b/aquadoggo/src/db/models/log.rs
@@ -98,7 +98,7 @@ impl Log {
     /// Operations are separated in different logs per document and author. This method checks if a
     /// log has already been registered for a document and author and returns its regarding log id
     /// or None.
-    pub async fn get(pool: &Pool, author: &Author, document_hash: &Hash) -> Result<Option<LogId>> {
+    pub async fn get(pool: &Pool, author: &Author, document_id: &Hash) -> Result<Option<LogId>> {
         let result = query_as::<_, LogId>(
             "
             SELECT
@@ -111,7 +111,7 @@ impl Log {
             ",
         )
         .bind(author)
-        .bind(document_hash)
+        .bind(document_id)
         .fetch_optional(pool)
         .await?;
 
@@ -125,11 +125,11 @@ impl Log {
     pub async fn find_document_log_id(
         pool: &Pool,
         author: &Author,
-        document_hash: Option<&Hash>,
+        document_id: Option<&Hash>,
     ) -> Result<LogId> {
         // Determine log_id for this document when a hash was given
-        let document_log_id = match document_hash {
-            Some(hash) => Log::get(pool, author, hash).await?,
+        let document_log_id = match document_id {
+            Some(id) => Log::get(pool, author, id).await?,
             None => None,
         };
 
--- a/aquadoggo/src/db/models/log.rs
+++ b/aquadoggo/src/db/models/log.rs
@@ -28,7 +28,7 @@ pub struct Log {
 }
 
 impl Log {
-    /// Register any new log_id for a document.
+    /// Register any new log_id for a document and author.
     ///
     /// The database will reject duplicate entries.
     pub async fn insert(
@@ -93,11 +93,12 @@ impl Log {
         Ok(next_log_id)
     }
 
-    /// Returns the registered log_id for a document.
+    /// Returns the registered log_id for a document identified by its hash.
     ///
-    /// Operations are separated in different logs per document and author. This method checks if a log
-    /// has already been registered for a document and returns its id.
-    pub async fn get(pool: &Pool, author: &Author, document: &Hash) -> Result<Option<LogId>> {
+    /// Operations are separated in different logs per document and author. This method checks if a
+    /// log has already been registered for a document and author and returns its regarding log id
+    /// or None.
+    pub async fn get(pool: &Pool, author: &Author, document_hash: &Hash) -> Result<Option<LogId>> {
         let result = query_as::<_, LogId>(
             "
             SELECT
@@ -110,7 +111,7 @@ impl Log {
             ",
         )
         .bind(author)
-        .bind(document)
+        .bind(document_hash)
         .fetch_optional(pool)
         .await?;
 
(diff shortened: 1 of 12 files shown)


aquadoggo/src/rpc/methods/publish_entry.rs
```diff
@@ -48,22 +45,10 @@ pub async fn publish_entry(
...
+    // log. An `UPDATE` or `DELETE` message is always stored in the same log that its original
     // `CREATE` message was stored in.
-    let document_log_id = match message.is_create() {
-        true => {
-            // A document is identified by the hash of its `CREATE` message
-            let document_hash = &params.entry_encoded.hash();
-            Log::find_document_log_id(&pool, &author, document_hash).await?
-        }
-        false => {
-            // An instance is identified by the hash of its previous operation
```
adzialocha, line 60: Maybe I'm misunderstanding this here but I stumbled upon this line and the regarding `get_log_id_by_instance` method. Is this the right definition of *instance*? I thought an Instance is a materialization of a Document and it is (still) identified by the hash of its `CREATE` message.

Also the implementation of `get_log_id_by_instance` returns a log id for any event hash which might not necessarily be one from an Instance (that event hash might just be holding an UPDATE Operation). I thought the name `get_log_id_by_entry` is more accurate here then.

Again, could be that I missed the crucial part here, let me know what you think!

adzialocha:
Meh, I'm seeing this error here:

```
thread 'rpc::methods::publish_entry::tests::publish_entry' panicked at 'assertion failed: `(left == right)`
  left: `"{\"error\":{\"code\":0,\"message\":\"Failed to decode entry: Could not decode signature: DecodeError\"},\"id\":1,\"jsonrpc\":\"2.0\"}"`,
 right: `"{\"id\":1,\"jsonrpc\":\"2.0\",\"result\":{\"entryHashBacklink\":\"00205a5f1de1b2fdb378a62305bf4b8a1ec137b6b6a008749d6252ba3011c58b266b\",\"entryHashSkiplink\":null,\"logId\":1,\"seqNum\":2}}"`', aquadoggo/src/rpc/methods/publish_entry.rs:245:9
 ```
 
The encoded entry looks okay ..
 
 ```
PublishEntryRequest { entry_encoded: EntrySigned("001c6b5c3222fef7c3f580a6bc34b3271f223dac7a9779f875c64daca1494069900101880020860e4ba776a71aa0ac13a0c9e3a1f21badb07ac48573542406582d5ea6942b8f11f24dfa5bc43a282487102646edde4c6089e24d5bd3ba0516b4b3ec302010b0b59e7725095185bf32a9f2e9e4d5687cc8d4be88eec8b3032b152a7cfab39f0b"), message_encoded: MessageEncoded("a466616374696f6e6663726561746566736368656d61784430303230623137376563316266323664666233623730313064343733653664343437313362323962373635623939633665363065636266616537343264653439363534336776657273696f6e01666669656c6473a16474657374a26474797065637374726576616c75656548656c6c6f") }
 ```

```
tag: 00
author: 1c6b5c3222fef7c3f580a6bc34b3271f223dac7a9779f875c64daca149406990
log_id: 01
seq_num: 01
payload_size: 88
payload_hash: 0020860e4ba776a71aa0ac13a0c9e3a1f21badb07ac48573542406582d5ea6942b8f
signature: 11f24dfa5bc43a282487102646edde4c6089e24d5bd3ba0516b4b3ec302010b0b59e7725095185bf32a9f2e9e4d5687cc8d4be88eec8b3032b152a7cfab39f0b
```

File: aquadoggo/src/rpc/methods/publish_entry.rs (hunk 1 above)
Comment (cafca) [reply aquadoggo/src/rpc/methods/publish_entry.rs:60]: Interesting! I think we should have discussed the definitions [here](https://github.com/p2panda/aquadoggo/issues/43) at the top together to talk about stuff like this.

> 1

An instance is a _specific_ materialisation of a document. A document can have many instances. That's why we can't identify all of those instances by the document id (hash of the create message).

> 2

I thought we can first implement everything (also materialisation, queries, etc) for single-writer because that's the base scenario rather than build everything at once. Then we have a full working stack and can "upgrade" it for the multi-writer scenario.

> 3

I think I confused myself here :D I followed the definition [here](https://github.com/p2panda/aquadoggo/issues/43) and I think reading _An instance is identified by the hash of the last operation applied._ my head went _look for the last->previous hash_. Going by the current operation's entry hash makes more sense.

> 4

Every operation creates a new instance. As all operations are stored in the same log in the single-writer scenario we can use this shortcut

Comment (adzialocha) [reply aquadoggo/src/rpc/methods/publish_entry.rs:60]: Thank you! I'll dive into it!

Would it be possible for you to sketch a client / node communication flow (similar to what we did once) for publishing new entries? Think it can help me to think about all of this closer to implementation.

Comment (adzialocha) [reply aquadoggo/src/rpc/methods/publish_entry.rs:60]: > my head went look for the last->previous hash. Going by the current operation's entry hash makes more sense.

`id` is giving you the document hash (former "instance id" this is where the name comes from), not the one of the current operation entry (this would be `hash`), maybe this is the confusion here?

If you would go by the current operation entry hash in that code you would get a new log id for every entry, regardless if they belong to the same document or not.

Comment (sandreae) [reply aquadoggo/src/rpc/methods/publish_entry.rs:60]: > 1
I had actually been imagining that the `Instance` would still be identified by the hash of the CREATE entry. I need to think more deeply though... Definitely worth discussing this together. Apart from that I had the same understanding of `Instance` @cafca

Comment (adzialocha) [reply aquadoggo/src/rpc/methods/publish_entry.rs:60]: > I think we should have discussed the definitions here at the top together to talk about stuff like this.

All the definitions make sense (except of maybe that an `Operation` is not encoded as an `Entry`, Operations are encoded as CBOR and always come *together* with an Entry which provides its integrity and authenticity, the hash of the attached Entry is also used to identify that Operation). Also I understood that `Instance` is still using the same hash as an identifier as the Document.

Instance doesn't seem to be a good word (I like "View" or "DocumentView") and also in general I find that concept important in theory but almost nothing a client will think about, especially when the hash is the same, its a detail I will not care about when querying the database via GraphQL (you could see it also as "I'm querying a Document in the eventually consistent state my Node has").

It gets harder since Instances are something which can only exist *after* materialization. I think this is also why I got confused seeing it here, since we don't have materialization implemented yet.

> I think I confused myself here :D I followed the definition here and I think reading An instance is identified by the hash of the last operation applied. my head went look for the last->previous hash. Going by the current operation's entry hash makes more sense.

More confusion analysis! We talked about "Log per Instance" for a long while *before* we changed the definition of an Instance. It should be called "Log per Document"! I could see that you changed the tests in a way where every newly created operation will lead to a new log id even though they belong to the same document, the code always uses a new hash and therefore a new log id.

Here maybe a small graphic on how I believe it is correct (let me know if that is false):

```
Document x:
log 1: [x] <- [y] <- [z] <- [l]              (key A)
log 2:          \- [i] <- [j] <- [k]         (key B)
```

This is how it is currently implemented after the "Instance per Log" changes:

```
log 1: [x]                                     (key A)
log 2:   \- [y]                                (key A)
log 3:        \- [z]                           (key A)
log 3:            \- [l]                       (key A)
```

Another thing which might have been confusing is the `id` in the `EncodedEntry` struct used in the code. It does not return the entry hash (this would be `hash`), but the document id! I guess we should call it `documentHash` or similar in the future?

> I thought we can first implement everything (also materialisation, queries, etc) for single-writer because that's the base scenario rather than build everything at once. Then we have a full working stack and can "upgrade" it for the multi-writer scenario.

It could make sense to already do *some* things in multi-writer, I actually think especially this flow should already cover it since materialization is relatively hard to think of without multi-writer (its fully designed and already supports multi-writer in the implementation). Multi-writer gets more complicated with Permissions etc., but just the materialization itself is already there :sparkles:

Comment (sandreae) [reply aquadoggo/src/rpc/methods/publish_entry.rs:60]: (nice graphics :heart: )

Comment (sandreae) [reply aquadoggo/src/rpc/methods/publish_entry.rs:60]: I really like this concept we have of being able to derive many `Instances` (I don't mind the naming, could be any of your suggestions too @adzialocha) from a single document. However I was thinking of it more as a conceptual/internal concept, and that clients would be making requests to `Documents` and getting back the most up-to-date `Instance` (materialisation). So actually they wouldn't know so much about the distinction and only one materialised result exists at a time.

File: aquadoggo/src/db/models/log.rs [hunk 2]
```diff
@@ -93,12 +93,12 @@ impl Log {
...
+    /// Returns the registered log_id for a document identified by its hash.
     ///
-    /// Operations are separated in different logs per document and author. This method checks if a log
-    /// has already been registered for a document and returns its id.
-    pub async fn get(pool: &Pool, author: &Author, document: &Hash) -> Result<Option<LogId>> {
-        // @TODO: Look up if system schema was used and return regarding log id
+    /// Operations are separated in different logs per document and author. This method checks if a
+    /// log has already been registered for a document and author and returns its regarding log id
+    /// or None.
+    pub async fn get(pool: &Pool, author: &Author, document_hash: &Hash) -> Result<Option<LogId>> {
```
sandreae, line 101: Following on from our absurd id naming jam... It's places like this where we could do with a naming convention. Just wondering how it would be reading this code relatively "cold", I would probably think `document_hash` was the actual hash of this document. Maybe whenever we are using a hash as an id we call it `document_hash_id` and when whenever it's the actual hash of that item, we call it `name_hash`?

Open to any other suggestions too though....

aquadoggo/src/rpc/methods/publish_entry.rs
```diff
@@ -41,52 +44,61 @@ pub async fn publish_entry(
...
+        // For any other operations which followed after creation we need to either walk the operation
+        // graph back to its `CREATE` operation or more easily look up the database since we keep track
+        // of all log ids and documents there.
+        //
+        // We can determine the used document hash by looking at what we know about the previous
+        // entry in this author's log.
+        //
+        // @TODO: This currently looks at the backlink, in the future we want to use
+        // "previousOperation", since in a multi-writer setting there might be no backlink for
+        // update operations! See: https://github.com/p2panda/aquadoggo/issues/49
```
sandreae, line 67: :+1:

aquadoggo/src/db/models/log.rs
```diff
@@ -120,43 +120,48 @@ impl Log {
...
+    /// Returns the related document for any entry.
     ///
-    /// Every instance is part of a document and, through that, associated with a specific log id
-    /// of its author. This method returns that log id by looking up the log that the instance's
-    /// last operation was stored in.
-    pub async fn get_log_id_by_instance(pool: &Pool, instance: &Hash) -> Result<Option<LogId>> {
-        let result = query_as::<_, LogId>(
+    /// Every entry is part of a document and, through that, associated with a specific log id used
+    /// by this document and author. This method returns that document by looking up the log id
+    /// that the document's last operation was stored in.
```
sandreae, line 149: Not sure if I understand this:

> looking up the log id that the document's last operation was stored in.

Where does the "last operation" come into this? As I understand this query, we're looking up the `document_id` from the log where the passed entry is stored.

sandreae (changes requested):
Great stuff, nice to be transitioning things over to our new world ;-p just some small clarification requested.

aquadoggo/src/db/models/log.rs
adzialocha: I like that, good idea!

aquadoggo/src/db/models/log.rs
adzialocha: Yes, thats confusing! I'll remove it :+1:

sandreae (approved):

PR #58, opened 2022-01-20
Add materialisation for a hard-coded sample schema

cafca:
Adds materialisation as part of publishing an entry, specifically for a hard-coded sample schema.

Closes #55 

## Todo

- [x] Fix binding of values in `write_document`
- [x] Make the upsert clause in `build_insert_query` generic
- [ ] Add tests

## Todos for other PRs

- [ ] Allow logs to start with an UPDATE operation if that operation has [_previous operations_](https://github.com/p2panda/aquadoggo/blob/2e6b2751ffec78ee5f3100ada3b0713d20e42bd2/aquadoggo/src/rpc/methods/publish_entry.rs#L65-L70): https://github.com/p2panda/aquadoggo/issues/49
- [ ] Update the p2panda-js api not to require previous operations (they should be cached/retrieved with next entry args
  - ~~[ ] Update next entry args to provide previous operations~~
  - [x] Update document view to contain `previous_operations` for all materialised documents 
    - [x] Store graph tip(s) in document tables
- [ ] Implement field inspection on `Schema` 
- [x] Update instance to allow getting its length without having to call `raw()` for that, remove `raw()`
- [ ] Store operations directly in the database for retrieval during materialisation (?)

## 📋 Checklist

- [ ] Revert p2panda-rs version in Cargo.toml
- [ ] Add tests that cover your changes
- [ ] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [ ] Link this PR to any issues it closes
- [ ] New files contain a SPDX license header

PR #61, opened 2022-01-27
Multi-writer document logs

sandreae:
Using `previous_operations` to determine causal connection between operations from different authors we can construct logs for multi-writer documents correctly.

This is a patch onto #58 

## 📋 Checklist


aquadoggo/src/rpc/methods/publish_entry.rs
```diff
@@ -132,8 +149,8 @@ pub async fn publish_entry(
         entry_backlink_bytes.as_deref(),
     )?;
 
-    // Register log in database when a new document is created
-    if operation.is_create() {
+    // Register log in database when entry is first item in a log.
+    if entry.seq_num().as_u64() == 1 {
```
adzialocha, line 153: You can use `entry.seq_num().is_first()` here :+1:

Issue #86, opened 2022-04-16
Communicate graceful shutdown to all services


adzialocha:
All services managed by `ServiceManager` (currently `TaskManager`) should have the option to get informed about a system-wide (graceful) shutdown to accordingly react to it.

Issue #84, opened 2022-04-16
Message bus for cross-service communication


adzialocha:
All services should be able to broadcast messages to each other on a common messaging bus. For now this would help to implement #86 but later it can also be used to send data across services, for example incoming entries for materialization etc.

PR #90, opened 2022-04-24
Service manager

adzialocha:
Introduces `ServiceManager` which orchestrates long-running concurrent services with a communication bus between them. Also it kindly informs all services about a shutdown and waits for them to stop before quitting the process for good.

The next step will be to a) remove `TaskManager` and replace it with this in the `Runtime` b) refactor the GraphQL API to fit the new "service" API and move it into the manager. c) Make use of the `axum` graceful shutdown feature.

Later we will also add the "Materializer" with the worker as a service here and use the communication bus to inform it about new, incoming entries from the GraphQL API. Much later the Replication service will also live here ..

Closes: #86 and #84 

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Introduce GraphQL endpoint [#81](https://github.com/p2panda/aquadoggo/pull/81)
 - Generic task queue with worker pool [#82](https://github.com/p2panda/aquadoggo/pull/82)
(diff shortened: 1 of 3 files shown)


sandreae (approved):
Looks really good to me. One thing I wondered is do we need to be able to get any more info from the `service_manager` about the state of the spawned services? I tried writing a test which had one service which completed it's job then stops, and I wanted to check the status of all services from the manager side, but this isn't possible now. Maybe it isn't something we need, but I just wondered. I approve of it greatly as it is now in any case, it has a nice simple API :-)

Issue #85, opened 2022-04-16
Rename task manager to service manager


adzialocha:
.. otherwise its too confusing with our new task queue

Issue #87, opened 2022-04-16
Better panic handling in all async task and services


adzialocha:
System critical panics inside any task workers or any async services should be handled in a controlled manner meaning that they should be unwinded outside of the async task and propagated up to the service manager / main thread, which then can accordingly react by either just crashing or attempting a "controlled" shutdown.

**Todo:**

- [x] Crash when something goes wrong in `ServiceManager` #92 
- [x] Crash when something goes wrong in worker `Factory`

PR #92, opened 2022-04-28
Refactor runtime and http service

adzialocha:
This PR ..

* [x] Moves `axum` http server into managed service
* [x] Uses graceful shutdown feature of `axum`
* [x] Replaces `TaskManager` with new `ServiceManager`
* [x] Makes the application shut down when a service returns an error or panics
* [x] Refactors file structure to prepare for more "service-centric" architecture

Closes: #85 and partly #87 

## 📋 Checklist


aquadoggo/src/server.rs
```diff
@@ -64,33 +35,40 @@ pub fn build_server(state: ApiState) -> Router {
...
+
+    axum::Server::try_bind(&http_address)?
+        .serve(build_server(context).into_make_service())
+        .with_graceful_shutdown(async {
+            signal.await.ok();
+        })
         .await?;
+
     Ok(())
 }
```
sandreae, line 55: Just checking, but this is the actual service function we add to the service manager, all services will be passed as functions with the same signature as this. Correct?

aquadoggo/src/service_manager.rs
```diff
@@ -62,29 +51,92 @@ where
...
+}
+
+// Service manager for orchestration of long-running concurrent processes.
+//
+// This manager offers a message bus between services for cross-service communication. It also
+// sends a shutdown signal to allow services to react to it gracefully.
+//
+// Stopped services (because of a panic, error or successful return) will send an exit signal which
+// can be subscribed to via the `on_exit` method. Usually stopped services indicate system failure
+// and it is recommended to stop the application when this events occurs.
```
sandreae, line 114: Cool, so this is the "stop everything if one service fails" pattern we just discussed in the previous PR?

aquadoggo/src/service_manager.rs
```diff
@@ -195,34 +282,77 @@ mod tests {
...
         });
 
         manager.shutdown().await;
 
         // Check if we received the message in all services
         assert_eq!(counter.load(Ordering::Relaxed), 5);
     }
+
+    #[tokio::test]
+    async fn on_exit() {
```
sandreae, line 317: :+1:

sandreae (approved):
Looks great, you already added the `exit_signal` :smile:

aquadoggo/src/server.rs
adzialocha: Yes exactly!

aquadoggo/src/service_manager.rs
adzialocha: Correct, the service manager in itself does not enforce this, but it gives you the option.

The aquadoggo cli `main.rs` uses that pattern and moves over to shutdown as soon as CTRL+C has been pressed or a service stopped aka the exit signal fired.

PR #102, opened 2022-05-09
Implement methods needed for replication defined in `EntryStore` traits

sandreae:
- implement new traits needed for replication defined here: https://github.com/p2panda/p2panda/pull/310

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/db/stores/entry.rs
+++ b/aquadoggo/src/db/stores/entry.rs
@@ -191,7 +191,7 @@ impl EntryStore<StorageEntry> for SqlStorage {
     /// Returns a result containing the entry wrapped in an option if it was found
     /// successfully. Returns None if the entry was not found in storage. Errors when
     /// a fatal storage error occured.
-    async fn entry_at_seq_num(
+    async fn get_entry_at_seq_num(
         &self,
         author: &Author,
         log_id: &LogId,
@@ -230,7 +230,7 @@ impl EntryStore<StorageEntry> for SqlStorage {
     /// Returns a result containing the latest log entry wrapped in an option if an
     /// entry was found. Returns None if the specified author and log could not be
     /// found in storage. Errors when a fatal storage error occured.
-    async fn latest_entry(
+    async fn get_latest_entry(
         &self,
         author: &Author,
         log_id: &LogId,
@@ -270,7 +270,10 @@ impl EntryStore<StorageEntry> for SqlStorage {
     /// Returns a result containing a vector of all entries which follow the passed
     /// schema (identified by it's `SchemaId`). If no entries exist, or the schema
     /// is not known by this node, then an empty vecot is returned.
-    async fn by_schema(&self, schema: &SchemaId) -> Result<Vec<StorageEntry>, EntryStorageError> {
+    async fn get_entries_by_schema(
+        &self,
+        schema: &SchemaId,
+    ) -> Result<Vec<StorageEntry>, EntryStorageError> {
         let entries = query_as::<_, EntryRow>(
             "
             SELECT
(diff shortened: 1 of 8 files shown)


aquadoggo/src/db/stores/entry.rs
```diff
@@ -15,6 +15,20 @@ use p2panda_rs::storage_provider::traits::{AsStorageEntry, EntryStore};
 use crate::db::models::entry::EntryRow;
 use crate::db::provider::SqlStorage;
 
+// Re-write once https://github.com/pietgeursen/lipmaa-link/pull/3 merged in lipma-link
```
pietgeursen, line 18: This is merged and released in `0.2.1` :tada:

pietgeursen: And also, yeah it's annoying the fact that technically lipmaa links can go all the way back to 0 but the bamboo spec starts link numbers at 1.

:thinking:

pietgeursen (approved):
Hell yeah! Excited for this!

aquadoggo/src/db/stores/entry.rs
sandreae: Yeh, I did a bit of a double take on that 😂

pietgeursen: I think I might add this to lipmaa link:

```rs
/// Get all the lipmaa link numbers back to some value
///
/// ```
/// use lipmaa_link::get_lipmaa_links_back_to;
///
/// let links = get_lipmaa_links_back_to(7, 1);
///
/// assert_eq!(links, vec![6, 5, 4, 1]);
///
/// ```
#[cfg(feature = "std")]
pub fn get_lipmaa_links_back_to(start: u64, lower_inclusive: u64) -> Vec<u64> {
    let mut path = Vec::new();

    let mut n = start;
    while n > lower_inclusive {
        n = lipmaa(n);
        path.push(n);
    }

    path
}
```

pietgeursen: Ok yeah, I think it's useful, published in `0.2.2`

sandreae: Nice! Thanks, I'll use that for sure.

sandreae:
Are you being blocked by this not being merged @pietgeursen ? Ideally I'd give it today for the others to review, but seeing as we're going into `development` anyway I think it would be ok to push ahead if that helps.

pietgeursen:
If I have more time tonight I'll just rebase on top of your changes.

So, no, not blocked 🙂

aquadoggo/src/db/stores/entry.rs
adzialocha: Oh yeah, I remember some one-off bug we had some time ago 😅

aquadoggo/src/db/stores/entry.rs
```diff
@@ -15,30 +17,70 @@ use p2panda_rs::storage_provider::traits::{AsStorageEntry, EntryStore};
 use crate::db::models::entry::EntryRow;
 use crate::db::provider::SqlStorage;
 
+#[derive(Debug, Clone, PartialEq)]
+
+pub struct StorageEntry {
+    entry_signed: EntrySigned,
+    operation_encoded: OperationEncoded,
+}
```
sandreae, line 25: I removed this previously because technically we don't need two structs (`EntryRow` and `StorageEntry`) to handle entries coming in and out of the db because the storage traits could be implemented directly on `EntryRow`. However, I decided to reintroduce it as I think it improves understanding of what each struct is for and we _do_ need different structs in most other stores, so it matches and will be recognisable across the repo.

cafca: We need to document this!

sandreae: Related issue here: https://github.com/p2panda/p2panda/issues/320

cafca (commented):
Branch was merged before review submitted so this is just half-way through

Issue #89, opened 2022-04-23
Custom implementations of `sqlx` `Encode` and `Decode` traits for db data types?


sandreae:
Would this be a neat way of replacing `EntryRow` which contains mostly untyped strings with a struct containing typed fields which can still be passed in and out of the dB?

Like this: 
https://github.com/launchbadge/sqlx/blob/b6e127561797fe9aababa24ec640275ecb9b42af/sqlx-core/src/sqlite/types/str.rs

adzialocha:
This is how we did it before in `p2panda_rs` but I didn't like having sqlx related code in the core library.

Is there a way to do it in aquadoggo?

sandreae:
Ah yeh, I remember that change. I wrote this meaning we would implement them in `aquadoggo`, this would work for `EntryRow` (as it is now) but yeh, not for everything...

sandreae:
I think what we had before was a little different from what I'm suggesting here actually.

Here's an example in the `sqlx` codebase: https://github.com/launchbadge/sqlx/blob/0b2844bf39128132aeccce2cfc02ec94472e074f/sqlx-core/src/decode.rs#L9-L59

I was thinking, for example, our `EntryRow` struct could contain proper `p2panda-rs` which get encoded as / decoded from strings in the background. Am I making sense here :rofl: ?

adzialocha:
I'm not sure if I understand correctly both your idea and the sqlx documentation, but I wonder if its not mixing up two different concepts?

`EntryRow` is a _row_ of a table, meaning it contains multiple columns of different _types_. So implementing `EntryRow` as a type does not make much sense?

adzialocha:
For writing traits for `p2panda` types it could just be that you can't implement traits for external structs. So even if you want to make it typed, you could only do it by creating a wrapper type first, ala:

`DoggoLogId(LogId)`

sandreae:
> EntryRow is a row of a table, meaning it contains multiple columns of different types. So implementing EntryRow as a type does not make much sense?

Yeh, I had mixed this up! ;-p

In any case, I had a little go at a custom implementation of `Decode` (see below). 

```rust

use std::error::Error;

use serde::{Deserialize, Serialize};
use sqlx::database::{Database, HasValueRef};
use sqlx::decode::Decode;
use sqlx::types::Type;

use p2panda_rs::identity::Author;

#[derive(Debug, Deserialize, Serialize)]
pub struct DoggoAuthor(Author);

[...]

PR #119, opened 2022-05-16
Add GraphQL client API

cafca:
Replaces the "ping" placeholder with a GraphQL API for p2panda clients.

- [x] Add a query endpoint for next entry arguments
- [ ] Add tests

## Next

- Interface with storage provider
- Add mutation for publishing entries
- Add query for requesting documents and/or schemas

## 📋 Checklist


sandreae (changes requested):
Hey, thanks for this, all the changes look good! However I think it makes sense to base this PR off the `development` branch. As there have been/will be a series of breaking changes (removing JSON rpc etc...) we are merging into there for now. In `development` you already have access to the initial storage provider implementation, so you would be able to use  `next_entry_args()` and `publish_entry()` and all other existing higher level storage methods there.

aquadoggo/src/graphql/client/root.rs
```diff
@@ -0,0 +1,131 @@
...
+        // Prepare database connection
+        let pool = ctx.data::<Pool>().map_err(|err| Error::from(err))?;
+        let provider = SqlStorage {
+            pool: pool.to_owned(),
+        };
+
+        provider
+            .get_entry_args(&args)
+            .await
+            .map_err(|err| Error::from(err))
```
cafca, line 53: In order to access the error here I had to require them to be Send + Sync via the storage provider trait definitions.

https://github.com/p2panda/p2panda/pull/324

cafca:
Oops, forgot something

cafca:
I added a utility in `u64_string.rs`: it let's you do something like this

```rust
pub struct EntryArgsResponse {
    #[serde(with = "log_id_string_serialisation")]
    pub log_id: LogId,

    #[serde(with = "seq_num_string_serialisation")]
    pub seq_num: SeqNum,

    pub backlink: Option<Hash>,

    pub skiplink: Option<Hash>,
}
```

and then `log_id` and `seq_num` are serialised as string values without having to change their type on this struct or having to implement a custom serialiser. I am not sure whether this will be helpful outside of testing, maybe it should also go into p2panda-rs. What do you think @sandreae ?

aquadoggo/src/graphql/client/u64_string.rs
```diff
@@ -0,0 +1,116 @@
...
+
+use core::fmt;
+
+use serde::de::{self, Unexpected, Visitor};
+
+/// Serialise log id as strings.
+///
+/// To be used as a parameter for Serde's `with` field attribute.
+#[allow(dead_code)]
+pub mod log_id_string_serialisation {
```
sandreae, line 11: This is very cool! It's similar to what I was trying to do in the context of storing and retrieving p2panda types from the db where they are mostly strings: https://github.com/p2panda/aquadoggo/issues/89

I prefer your approach as it seems more generally useful. I don't know if it is a drop-in solution to for the db stuff though. Definitely let's keep an eye on it and see if we want to implement it in `p2panda-rs` and then try to use for the db models.

sandreae (approved):
This all looks good! Great to have the client GraphQL being introduced and finally completely removing JSON RPC. 

Interesting to see what we can do with the string serialisation definitions too :+1:

PR #123, opened 2022-05-17
Entry store updates review

cafca:
Describe your changes here.

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/db/stores/log.rs
+++ b/aquadoggo/src/db/stores/log.rs
@@ -1,52 +1,61 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-use std::str::FromStr;
-
 use async_trait::async_trait;
+use sqlx::{query, query_scalar};
+
 use p2panda_rs::document::DocumentId;
 use p2panda_rs::entry::LogId;
 use p2panda_rs::identity::Author;
 use p2panda_rs::schema::SchemaId;
 use p2panda_rs::storage_provider::errors::LogStorageError;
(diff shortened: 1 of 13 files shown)


aquadoggo/src/db/stores/log.rs
```diff
@@ -11,7 +11,10 @@ use p2panda_rs::storage_provider::errors::LogStorageError;
 use p2panda_rs::storage_provider::traits::{AsStorageLog, LogStore};
 
 use crate::db::provider::SqlStorage;
-
+/// Represents a stored log in a way that is conducive to implementing storage traits.
+///
+/// In the case of `LogRow`, the traits could have been implemented directly on it but this
+/// structure was chosen to align with the way the other stores are structured.
```
cafca, line 17: @sandreae is this correct?

sandreae: Fun fact: we actually never retrieve a log from storage.... only the log id... so that's even more redundancy :rofl:

sandreae: I don't know how best to write that, saying that the later is "laid out like the trait methods params"  seems off to me. The reason for each struct is:

1) `sqlx` can infer the type of retrieved returned query if there is a struct with matching rows, that's `LogRow`
2) Our `StorageProvider` trait requires each sub-store to also implement a trait relating to each stores primary data type, that's `StorageLog`

In this case we could make `1)` and `2)` the same struct, but currently choose not to for consistency across storage modules.

PR #124, opened 2022-05-17
Casted u64 index

### LINKED ISSUE #122: Add casted index for u64 fields that are stored as strings
Some u64 values are stored as strings because SQLite doesn't store u64. In some queries such as `SqlStorage::get_paginated_log_entries` these values are cast as numeric values.

We should add an index on those casted values so they don't have to be generated for each query.

cafca:
Closes #122 

I checked that the index is actually used with 

```sql
EXPLAIN QUERY PLAN SELECT 
                author,
                entry_bytes,
                entry_hash,
                log_id,
                payload_bytes,
                payload_hash,
                seq_num
            FROM
                entries
            ORDER BY
                CAST(seq_num AS NUMERIC) DESC
            LIMIT
                1
```

=> 

```sql
SCAN TABLE entries USING INDEX idx_entries_by_seq_num
```

## 📋 Checklist

- [ ] ~~Add tests that cover your changes~~


sandreae (approved):
Brilliant! Thanks for spotting these SQLite improvements :pray:

PR #130, opened 2022-05-19
Implement `SchemaStore`

sandreae:
I went for an approach where we don't use specialist tables for schema. As we already have all documents and views stored in the db, we can use these to retrieve and construct the schema. As we plan to keep them in memory, I think this approach is fine. If at a later date we want to retain some other data related to schema, or for any other reason, we can create a table. For now i didn't think there was any need.

This store pattern should be familiar by now :rofl: 

## 📋 Checklist


adzialocha (approved):
This looks very clear and nice! Yay!

PR #132, opened 2022-05-19
Add graphql endpoint for publishing entries

cafca:
Reopened from #128 (which was closed by buggy Github).

Todo

- [x] Implement endpoint
- [x] Happy path test
- [ ] Sad path test

## 📋 Checklist


aquadoggo/src/graphql/client/mutation.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        let query = r#"
+            mutation TestPublishEntry($entryEncoded: String!, $operationEncoded: String!) {
+                publishEntry(entryEncoded: $entryEncoded, operationEncoded: $operationEncoded) {
+                    logId,
+                    seqNum,
+                    backlink,
+                    skiplink
+                }
+            }"#;
```
sandreae, line 87: This is just me not knowing anything about GraphQL, but where does this `TestPublishEntry` come from? Or rather, what is it?

sandreae (approved):
Looks great! Thanks :+1: 

I just had one question because I don't know GQL at all, but I'm happy for this to merge.

aquadoggo/src/graphql/client/mutation.rs
cafca: All of this is also new to me! We have been using the shorthand syntax so far but usually I think you give a name to requests you can make multiple requests to the same endpoint with one query.

https://graphql.org/learn/queries/#operation-name

Issue #134, opened 2022-05-26
Move storage provider into context


adzialocha:
Currently we create a new storage provider instance on each GraphQL request, we could instead initialize it once and move it into the shared context.

PR #135, opened 2022-05-27
Move `SqlStorage` into shared `Context`

adzialocha:
This moves the `SqlStorage` instance into the shared `Context`, so we don't need to create a new one on every GraphQL request (and other future methods).

Note: It is still possible to access the underlying connection pool object when needed as it is publicly exported inside the `SqlStorage`.

Closes: #134 

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/graphql/client/query.rs
+++ b/aquadoggo/src/graphql/client/query.rs
@@ -6,12 +6,10 @@ use p2panda_rs::identity::Author;
 use p2panda_rs::storage_provider::traits::StorageProvider;
 
 use crate::db::provider::SqlStorage;
-use crate::db::Pool;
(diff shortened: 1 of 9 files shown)


aquadoggo/src/graphql/client/query.rs
```diff
@@ -41,13 +39,9 @@ impl Query {
...
-        // Prepare database connection
-        let pool = ctx.data::<Pool>()?;
-        let provider = SqlStorage {
-            pool: pool.to_owned(),
-        };
-
-        provider.get_entry_args(&args).await.map_err(Error::from)
+        // Load and return next entry arguments
+        let store = ctx.data::<SqlStorage>()?;
+        store.get_entry_args(&args).await.map_err(Error::from)
```
cafca, line 44: Much better! 🌷

Issue #127, opened 2022-05-18
High idle CPU utilization


cafca:
With idle workers, aquadoggo on `development` uses ~400% CPU for me.

adzialocha:
I can imagine this is due to our loop implementation where every cycle wakes the runtime even when the queue is empty (`loop` and `yield_now`). We added this to not starve the runtime with an infinite loop but it's not Rust async best-practice.

Probably this can be solved with a scheduler logic (future stream) which only wakes the runtime when there is a task waiting in the queue, maybe even combined with a round robin strategy checking that all queues are equally considered.

adzialocha:
Enjoyed reading this article which probably solved a very similar issue of scheduling: https://smb374.github.io/an-adventure-in-rust-s-async-runtime-model.html

PR #136, opened 2022-05-27
Fix high CPU usage of idle workers

adzialocha:
This uses [`deadqueue`](https://github.com/bikeshedder/deadqueue), a FIFO-queue implementation wrapping `crossbeam_queue::SegQueue`. It extends it with a future on `pop` which uses a nice [semaphore](https://github.com/bikeshedder/deadqueue/blob/master/src/unlimited.rs#L34) logic internally allowing us to `await` on it. This gives us exactly what we need to not naively unblock the async task and retry the queue manually on _every_ loop cycle, as there was otherwise no future to wait for before. This led originally to unnecessary waking of the runtime even when there was nothing to do, wasting CPU resources. With the new `await` / future on the queue we have a chance now to do something else in the async runtime and come back as soon as there is something to do again.

Closes: #127 

## 📋 Checklist


adzialocha:
Doggo not working so hard anymore :smiling_face_with_tear: 

screenshot-2022-05-27T20:07:21

sandreae (approved):
Looks like a very neat solution :+1: and thanks for tidying up the PR hash prefixes ;-p

Issue #72, opened 2022-03-04
Dynamic GraphQL schema introspection


adzialocha:

cafca:
I think this should be closed by #141 and isn't closed already?

adzialocha:
Yeah, thats true, but also mentioning here was a little redundant as with the new approach we get it "for free" without needing to "do" anything. So I thought its not really a "todo" anymore :-D

adzialocha:
I had the feeling its a ticket which rather confuses us (what! WHAT DO I HAVE TO IMPLEMENT?!)

cafca:
No problem :) It's just a quite distinct subtask within #141 so it kind of made sense to me to have it as a ticket.

adzialocha:
> No problem :) It's just a quite distinct subtask within #141 so it kind of made sense to me to have it as a ticket.

Yeah, I get that as well! Feel free to reopen :+1:

Issue #64, opened 2022-01-29
Dynamically generate GraphQL API from application schemas


adzialocha:
# Development

## Stage 1

- Using the StorageProvider
- No cute filter and queries

## Stage 2

- Use sql queries in order to provide ordering and filtering by fields

## Stage 3

- Bring back the advanced filtering and sql queries into the storage provider

Outline for GraphQL API:

```
events_0020...38830
    parameter
      id
      view_id
      
    returns an object representing a document view
        fields
            ...depending on the schema
        meta
            id: String!
            view_id: String!
            authors: String[]
            deleted: Boolean
            edited: Boolean
            operations
        reverse_relations [tbd]
            [reverse_field_name]
        
all_events_0020...38830
    parameter
        orderBy
        orderDirection
        first
        skip  
        
    returns an array of objects
```

PR #141, opened 2022-06-01
Dynamic GraphQL document API

cafca:
## Main Course 🍲

- Generates dynamic GraphQL schemas supporting introspection
- Handles requests against these schemas

## Side Dishes 🍚

- Adds `context` field to `TestDatabase`
- Adds `add_document()` and `add_schema()` helpers on `TestDatabase`

Depends on: #166 and https://github.com/p2panda/p2panda/pull/416

Closing #64, #72, https://github.com/p2panda/p2panda/issues/330

## 📋 Checklist

- [x] All spec changes reflected in handbook
  - https://github.com/p2panda/handbook/pull/198
- [x] Merge https://github.com/p2panda/p2panda/pull/416
- [x] Merge #204 

## Next

- Compute complexity of queries to prevent expensive recursive queries
- Add caching directives for fields that hit the database (debate?)
- https://github.com/p2panda/aquadoggo/issues/222
- https://github.com/p2panda/aquadoggo/issues/227


adzialocha:
Wow! This is cool! Of course its lame we can't get access to the pool (https://github.com/async-graphql/async-graphql/discussions/930) but your idea with reading from a .tmp file works.

I can see the following flow here with this approach, please correct me if thats wrong:

- Node admin has a list of schema ids it wants to support (for example via command line args, env vars etc.)
- Node gets started, the .tmp file gets generated based on what the node already knows about them (incomplete schemas get ignored)
- Node needs to be restarted again to build the "dynamic" schema from the just-generated file
- Node is ready to support these schemas (given that they were in the database at that time), if some schema definitions arrive later, the process needs to be repeated until the node fully supports the schema

cafca:
Roughly that's what I thought as well, yes. If the node already has all wanted schemas materialised I would hope there is no restart needed though because the temp file can be generated before the graphql schema is initialised.

Do you have an idea for a way to restart the node, or even better the http server specifically, once a schema is materialised?

adzialocha:
> Do you have an idea for a way to restart the node, or even better the http server specifically, once a schema is materialised?

Ja, there could be a way, but wouldn't that interrupt connection and cause errors for every client talking to the node whenever there is a new schema arriving? I'd rather would have a manually controlled shutdown (at a "good time", with announcement etc.). Also we start introducing more hacky complexity out of a hack which I would try to avoid.

cafca:
Yes, of course, dropping connections wouldn't be nice. I thought that `graceful_shutdown` thing means that the server waits until all in-flight requests are finished before shutting down and something like that could also be done with restarting.

adzialocha:
> Yes, of course, dropping connections wouldn't be nice. I thought that `graceful_shutdown` thing means that the server waits until all in-flight requests are finished before shutting down and something like that could also be done with restarting.

Thats true, we could make sure the services wind down nicely.

> If the server is just unavailable for new connections for a second while it restarts I think that wouldn't be so bad. Clients could still retry any requests or no?

The problem is rather what happens in the moment when they are down and before they come up again. Sure, we can ask clients kindly to retry but most of them will probably just error out (or did your browser retry requesting a page on an error)? :cat2:

adzialocha:
Random thought I had the other day: Since we're getting the schema file statically during startup I was wondering if that could be achieved with the `lazy_static` macro, and then I was wondering if we could even make the database query inside of it? This would require `async` and this sounds tricky, but this looks promising: https://crates.io/crates/async_once

adzialocha:
I've tried to bring this branch back to the latest state of `development` as the context organisation changed a lot. I needed to do some "deeper cuts" into your work @cafca as I didn't wanted to combine my merging attempt with large rewrites, hope that was okay. This is what I needed to do:

1. The branch you worked off from `p2panda-rs` changed since last time, so there are some breaking API changes now I didn't handle .. `SchemaService` has a couple of holes now because of this (I've commented the sections out). I guess that can be sorted out as soon as that PR is merged :+1:
2. Making the storage provider of `SchemaService` generic made the merging trickier, I've just changed its type to `SqlStorage` for now

If all of this sucks, of course feel free to roll back to the commits before my merge :+1:

adzialocha:
Ha! New idea around that "restarting" problem.

How about making `schemas` in the HTTP context an actual array, with the last item being the "latest" schema? It would be selected here: https://github.com/p2panda/aquadoggo/blob/development/aquadoggo/src/http/api.rs#L20

When the schema provider added a new schema to the node, it will trigger `build_root_schema` and add it to the array of the HTTP context. Every future request will automatically be executed on that new GraphQL schema, while pending / old ones will be finished with the old?

I can try to draft a POC and see if that works ..

cafca:
Wow, that sounds amazing! It could work and even has a nice handover when schemas change.

adzialocha:
Alright, I think this should do the job now :partying_face: I guess the next thing would be to get the https://github.com/p2panda/aquadoggo/pull/166 PR merged, so we can integrate the `SchemaProvider` properly here (the `SchemaService` is outdated) and continue making these dynamic queries _really_ nice. Exciting!

aquadoggo/src/graphql/schema.rs
```diff
@@ -39,3 +47,143 @@ pub fn build_root_schema(store: SqlStorage, tx: ServiceSender) -> RootSchema {
...
+/// static access to the database when building `async_graphql` types.
+///
+/// With this little workaround we are still able to make it work! We load the p2panda schemas from
+/// the database and serialise them into a JSON file. When `async_graphql` builds the GraphQL
+/// schema we can load this file statically to build the schemas dynamically based on the file's
+/// content.
+async fn build_schema_with_workaround(shared: GraphQLSharedData) -> RootSchema {
+    // Store temporary JSON file with all serialised application schemas from database.
+    //
+    // @TODO: We could try to still access the database by creating a static interface to it?
```
adzialocha, line 68: :point_up: I'd like to explore this for a short while, just to see how far we can get.

aquadoggo/src/graphql/schema.rs
```diff
@@ -39,3 +47,143 @@ pub fn build_root_schema(store: SqlStorage, tx: ServiceSender) -> RootSchema {
...
+
+        manager
+    }
+
+    /// Subscribes to `SchemaService` for newly added schemas.
+    ///
+    /// This spawns a task which listens to new p2panda schemas to accordingly build a GraphQL
+    /// schema which will be added to the list.
+    // @TODO: This manager does not "clean up" outdated schemas yet, they will just be appended to
+    // an ever-growing list.
```
adzialocha, line 142: :point_up: Thats of course a potential source of ever-growing memory the longer the node runs, but I don't see a need to fix this in this PR.

I believe the mutex protects us from removing schemas from the list _while_ they are being executed somewhere else due to a pending request. So it should be safe to just remove items?

aquadoggo/src/schema_service/schema_service.rs
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+// @TODO: The API changed here in `p2panda_rs`
```
adzialocha, line 3: :point_up: This whole thing gets replaced by the `SchemaProvider`, which is developed here: https://github.com/p2panda/aquadoggo/pull/166 - so no need to worry too much about this file.

aquadoggo/src/schema_service/schema_service.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /// Registers a new schema in the provider.
+    pub async fn update(&self, schema: &Schema) -> Result<(), SchemaServiceError> {
+        // @TODO: Add it to the store
+
+        // Inform subscribers about new schema
+        if self.tx.send(schema.id().to_owned()).is_err() {
+            // Do nothing here, as we don't mind if there are no subscribers
+        }
```
adzialocha, line 54: :point_up: This is important and needs to be added to the new `SchemaProvider` as soon as it gets integrated here. It will make sure that the GraphQL schema manager will be notified as soon as we've updated a schema, so it can build a new GraphQL root schema.

aquadoggo/src/schema_service/temp_file.rs
```diff
@@ -0,0 +1,77 @@
...
+        let file = File::open(path).unwrap();
+        let reader = BufReader::new(file);
+        serde_json::from_reader(reader).unwrap()
+    }
+
+    /// Restore data and pin it to memory for the rest of the program's runtime.
+    ///
+    /// **Attention**: dropping the returned reference will leak memory.
+    pub fn load_static(path: &str) -> &'static Data {
+        Box::leak(Box::new(Self::load(path)))
```
adzialocha, line 34: I wonder if we can make this whole "file thing" a static object living throughout the program by using `lazy_static` or the more fresh [`once_cell`](https://docs.rs/once_cell/latest/once_cell/index.html)

aquadoggo/src/graphql/schema.rs
cafca: What if the schema was used in a way that involves multiple locks during one process? Like a subscription request, could that involve a series of locks?

cafca:
I clumsily merged the schema provider branch :') I also updated the api of schema provider with the broadcast channel but I think it's not actually changing anything? Something's not right. I tried with these graphql queries for creating a schema field definition and then a schema definition. The schema is added to the schema provider but the graphql schema doesn't change.

schema field definition
```graphql
mutation put {
  publishEntry(
    entryEncoded: "002f8e50c2ede6d936ecc3144187ff1c273808185cfbc5ff3d3748d1ff7353fc96010179002026dcc8e0d61fb74ba22fa9d27ec50bc70002bd0970d0ac9b4b0e23d2657c0e2b810ef0b289b60e844da36bbe44a9849f280bb55caa9b403951d3dd6eb354fadfc73136fc4bfdae531a80cb7b92d07ffc4d1b1c4de3433efdcf534ad41d02fc08",
    operationEncoded: "a466616374696f6e6663726561746566736368656d61781a736368656d615f6669656c645f646566696e6974696f6e5f76316776657273696f6e01666669656c6473a2646e616d65a26474797065637374726576616c7565676d6573736167656474797065a26474797065637374726576616c756563737472"
  ) {
    logId,
    seqNum,
    backlink,
    backlink,
    skiplink
  }
}
```

schema definition
```graphql
mutation put {
  publishEntry(
    entryEncoded: "002f8e50c2ede6d936ecc3144187ff1c273808185cfbc5ff3d3748d1ff7353fc960201f8f900207c342547b87a0c156f3529761f87a4161f04235a3c3b412fe22c7dcd32cfd2b23dc63c82e074a8fc09c397d8a5ef721b6fc74213e49c6a4379222b9b366c3bc56f8d9525b8c3fc6eeb248a2f851d0ba9924b3919e1a64cdb4c50b37e48f5ba0c",
    operationEncoded: "a466616374696f6e6663726561746566736368656d6174736368656d615f646566696e6974696f6e5f76316776657273696f6e01666669656c6473a36b6465736372697074696f6ea26474797065637374726576616c75656e53656e64206d657373616765732e666669656c6473a264747970657470696e6e65645f72656c6174696f6e5f6c6973746576616c7565818178443030323065343836373239383235623039666330666261313036653833343730306334356166663737346633636266313332613235313461616133366133636333346663646e616d65a26474797065637374726576616c75656c636861745f6d657373616765"
  ) {
    logId,
    seqNum,
    backlink,
    backlink,
    skiplink
  }
}
```

adzialocha:
> Something's not right

I've just made some tests and found out that the sender fails due to no subscribers, will investigate :+1:

But wow, this is already sooo cooooool.

adzialocha:
Found the issue, the `ServiceProvider` existed twice (once created in `Node` and once created in `http_service`), I've made sure now that there is only one and it works! :balloon:

Screenshot 2022-06-23 at 11-39-32 Playground - http __localhost 2020_graphql

aquadoggo/src/graphql/schema.rs
adzialocha: Yeah, subscriptions will need also some special interface between but I don't think its an issue. We will just have to do something similar as we do with the `GraphQLSchemaManager`, aka, something which abstracts away that there is actually a changing schema behind.

cafca:
Wow! I have to check that out.

cafca:
I am changing the temp file helper to use the default path returned by our `Configuration`. I guess that's better than just using the working directory. However, I saw some errors in `cargo t` that I suspect happen because tests are executed simultaneously and end up writing to the same temp file during startup.

Evidence, lol :D

Screenshot 2022-06-22 at 23 56 41

adzialocha:
> I suspect happen because tests are executed simultaneously and end up writing to the same temp file during startup.

Oh wow, thats exactly the problem I have with the PostgreSQL CI tests. I guess we have to give the `.tmp` file a random name to allow concurrent tests?

adzialocha:
Also, since you already got the `Configuration` into the temp-file thingy, why not talk to the database? :-D

adzialocha:
I've had a larger deep-dive into this whole topic of making it work with `async-graphql`, just to understand better what is required. This led to a few realizations, both in my brain and in code:

* I've added a whole bunch of documentation explaining our approach
* We do not need to create temporary files! We can simply create a global schema provider where we get access to from all directions (including the inner `async-graphql` methods). I've refactored that bit and this creates now a `Vec<Schema>` wrapped in `once_cell::sync::Lazy` which is the cooler variant of `lazy_static`. Nice side-effect: The tests are passing since there is no race-conditions overwriting the temporary files.
* `Box::leak` _can_ cause a memory leak, so far we don't as the created data actually really lives as long as the program. As soon as we start "cleaning up" old GraphQL schemas we will have to take care of that though. I've explored a little bit on how to do this and prepared a `StaticLeak` type which might help us with this problem. It uses `unsafe` code, but again `Box::leak` is not safer anyhow .. Also I've added a bunch of notes and warnings in all places I think which might become dangerous as soon as we start touching them again. I don't think we need to tackle automatic schema clean-up in this PR, it will need some more time and care to do this right, this is just a preparation and precaution-measure for the future. For now we will be fine with a forever-growing schema store (I guess a "memory leak" in another way lol).

aquadoggo/src/schema/static_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+        assert_eq!(*value, 100);
+
+        // We drop the struct before the program ended, this will free the used memory for the
+        // struct and the inner value! We do not have a memory leak!
+        drop(managed_leak);
+
+        // Now it gets bad .. `value` still exists in scope and the borrow checker doesn't warn us
+        // that it got dropped because it still assumes it has 'static lifetime. But actually it is
+        // gone and you will probably not find the value `100` here anymore ..
+        assert_ne!(*value, 100);
```
cafca, line 190: This test is pretty cool :D 

But it fails sometimes for me

```
running 1 test
thread 'schema::static_provider::tests::drops_static_value' panicked at 'assertion failed: `(left != right)`
  left: `100`,
 right: `100`', aquadoggo/src/schema/static_provider.rs:194:9
```

What do?

cafca: Could the problem be that `100` is not placed in allocated memory because it's just a teensy integer that implements `Copy`? If I replace the value with `"Hello".to_string()` the test fails with

```
error: test failed, to rerun pass '-p aquadoggo --lib'

Caused by:
  process didn't exit successfully: `/Users/work/code/zoo/aquadoggo/target/debug/deps/aquadoggo-dced19c1730dfd87 'schema::static_provider::tests::drops_static_value' --exact --nocapture` (signal: 11, SIGSEGV: invalid memory reference)
```

Which seems appropriate!

adzialocha: Haha, lol, maybe your computer manages memory different? Would need to read up on that, but I doubt it has something to do with the Copy trait (I guess it's even lower since it works on Linux) 🤔

aquadoggo/src/graphql/client/dynamic_types/document_meta_type.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::graphql::client::dynamic_types::utils::{metafield, metaobject};
+use crate::graphql::scalars::{
+    DocumentId as DocumentIdScalar, DocumentViewId as DocumentViewIdScalar,
+};
+
+/// Name of the field for accessing the document's id.
+const DOCUMENT_ID_FIELD: &str = "documentId";
+
+/// Name of the field for accessing the document's view id.
+const DOCUMENT_VIEW_ID_FIELD: &str = "documentViewId";
```
adzialocha, line 16: I like this being explicit but we had some ideas to change the API to `nextArgs(publicKey, viewId`), so it would be good to find some consistent naming for document view ids across the API. `viewId` is also nice, its so short?

cafca: I like that!

aquadoggo/src/graphql/client/dynamic_query.rs
```diff
@@ -0,0 +1,334 @@
...
+use crate::db::traits::DocumentStore;
+use crate::graphql::client::dynamic_types::DocumentMetaType;
+use crate::graphql::scalars::{
+    DocumentId as DocumentIdScalar, DocumentViewId as DocumentViewIdScalar,
+};
+use crate::schema::SchemaProvider;
+
+use crate::graphql::client::utils::validate_view_matches_schema;
+
+/// Container object that injects registered p2panda schemas when it is added to a GraphQL schema.
```
cafca, line 29: note to self: outdated docstring

aquadoggo/src/schema/static_schema_provider.rs
```diff
@@ -0,0 +1,194 @@
...
+            Schema::new(&schema_id, "Test schema", vec![("name", FieldType::String)]).unwrap();
+        let schemas = vec![schema];
+        save_static_schemas(&schemas);
+
+        // Load from global storage
+        let from_store = load_static_schemas();
+        assert_eq!(&schemas, from_store);
+    }
+
+    // #[test]
```
adzialocha, line 176: I'm happy to remove this for now or add a `@TODO`, up to you :+1:

adzialocha: We probably need more time than we have to figure this out and we don't even use it yet

cafca: Interesting: I recently updated Rust and now this seems to pass. I want to try whether the error comes back when I go back to an old Rust version.

cafca: Ok, I remove it.

cafca:
Hmm, I wonder why gh clippy doesn't know [this](https://rust-lang.github.io/rust-clippy/master/index.html#explicit_auto_deref) rule that was added in 1.60 even though the test runner is on 1.61.

aquadoggo/src/materializer/worker.rs
```diff
@@ -495,10 +495,8 @@ where
                         }
                     }
 
-                    // Trigger status update when successful
-                    if result.is_ok() {
-                        on_complete(item.input());
-                    }
+                    // Trigger removing the task from the task store
+                    on_complete(item.input());
```
cafca, line 499: I made this change so tasks that fail silently don't get rescheduled after a node restart.

aquadoggo/src/graphql/client/dynamic_types/dynamic_output_type.rs
```diff
@@ -0,0 +1,132 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! Registers GraphQL types for all [schemas][`Schema`] available in the schema provider.
+//!
+//! `async_graphql` doesn't provide an API for registering types that don't correspond to
+//! any Rust type. This module uses undocumented, internal functionality of `async_graphql` to
+//! circumvent this restriction. By implementing `OutputType` for [`DynamicQuery`] we are given
+//! mutable access to the type registry and can insert types into it.
+
```
adzialocha, line 9: Remove newline

aquadoggo/src/graphql/client/dynamic_types/dynamic_output_type.rs
```diff
@@ -0,0 +1,132 @@
...
+use std::borrow::Cow;
+
+use crate::graphql::scalars::{
```
adzialocha, line 12: Fix import order

aquadoggo/src/graphql/client/dynamic_types/mod.rs
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! This module contains dynamic type definitions for the GraphQL client api.
+//!
+//! All dynamic type definitions are inserted from the `OutputType` implementation in the
+//! [`dynamic_output_type`] module.
+
```
adzialocha, line 7: Remove newline

aquadoggo/src/graphql/client/dynamic_types/tests.rs
```diff
@@ -0,0 +1,270 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! Test correct generation of output schema.
+
```
adzialocha, line 4: Remove newline

aquadoggo/src/graphql/client/tests.rs
```diff
@@ -0,0 +1,211 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! Integration tests for dynamic graphql schema generation and query resolution.
+
```
adzialocha, line 4: Remove newline

aquadoggo/src/graphql/client/static_types/next_entry_arguments_type.rs
```diff
@@ -10,7 +10,7 @@ use crate::graphql::scalars;
 
 /// Arguments required to sign and encode the next entry for an author.
 #[derive(SimpleObject, Serialize, Deserialize, Debug, Eq, PartialEq)]
-pub struct NextEntryArguments {
+pub struct NextEntryArgumentsType {
```
adzialocha, line 13: You mentioned that on the chat, I think it's cool to rename all scalars to contain a "Scalar" suffix ("Type" is too abstract).

I'm also fine with _not_ doing that. So far I circumvented confusion by only importimg scalars together with the parent module ala `scalars::NextEntryArguments` which worked for me, but I understand that this is very subjective 😅

aquadoggo/src/db/stores/test_utils.rs
```diff
@@ -380,10 +392,97 @@ pub fn test_db(
...
+    pub context: Context,
     pub store: SqlStorage,
     pub test_data: TestData,
 }
 
+impl TestDatabase {
+    /// Publish a document and materialise it in the store.
+    ///
+    /// Also runs dependency task for document.
+    pub async fn add_document(
```
sandreae, line 404: These are really cool methods, I like how they work. 

I might like to also explore a flow where we use existing `send_to_store()` or db config with `populate()` functions for inserting entries and operations and then a new function called `materialise()` which just materialises everything in the store. I like the pattern of _not_ adding any new methods to the db store and instead modifying with helpers. I think this gives nice separation between what is possible with the store alone and what is extended test functionality.

For `add_schema` there could be  a helper function which takes a store and fields and then inserts the needed operations internally. Quite similar to what's here, just not on the test db.

I'm very happy to raise this is an issue and discus further though.

aquadoggo/src/graphql/client/dynamic_query.rs
```diff
@@ -0,0 +1,556 @@
...
+            Err(_) => Ok(None),
+        }
+    }
+}
+
+impl DynamicQuery {
+    /// Returns a single document as a GraphQL value.
+    async fn query_single(
+        &self,
+        schema_id: &SchemaId,
```
sandreae, line 73: Where does this `SchemaId` come from? The query string right? Sure it's just some `async_graphql` magic...

aquadoggo/src/graphql/client/dynamic_query.rs
```diff
@@ -0,0 +1,556 @@
...
+                            format!("Could not parse `id` argument: {}", err.message),
+                            Some(pos),
+                        )
+                    })?,
+                    ctx,
+                    ctx.field().selection_set().collect(),
+                    Some(schema_id),
+                )
+                .await?,
+            )),
```
sandreae, line 116: I find these inline/nested params a little hard to read/grok, but maybe that's just cos I'm less familiar with the patterns used here.

aquadoggo/src/graphql/client/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+use async_graphql::{Pos, ServerError, ServerResult};
+use p2panda_rs::document::DocumentViewId;
+use p2panda_rs::schema::SchemaId;
+
+use crate::db::provider::SqlStorage;
+
+/// Validate that the given view matches the given schema.
+///
+/// Panics if the view id does not exist in the store.
+pub async fn validate_view_matches_schema(
```
sandreae, line 12: Should/will this validation be happening at a different point in our stack later on?

aquadoggo/src/graphql/client/tests.rs
```diff
@@ -0,0 +1,210 @@
...
+                        }
+                    }],
+                    "by_pinned_relation_list": [{
+                        "fields": {
+                            "it_works": true
+                        }
+                    }]
+                }
+            }
+        });
```
sandreae, line 206: Whoop!

sandreae (approved):
Wow, this is an insanely magical PR. To be honest, I'm not really able to get my head around the `async_graphql` trickery you managed to work... but I have used the API a bit and I think it's great.

With that in mind, I think this is a fairy surface level review, but I will keep in mind I want to poke around here more in the coming months :+1:

Thanks so much for figuring all this out, it's awesome :star_struck:

aquadoggo/src/graphql/client/utils.rs
cafca: Yes please!

aquadoggo/src/graphql/client/dynamic_query.rs
adzialocha: I think its literally just a string from the GraphQL query (like `venue_0020...`) and it gets split at position 4 (after `all_`) and that string is used to parse it (`SchemaId::from_str`)

aquadoggo/src/schema/static_schema_provider.rs
```diff
@@ -0,0 +1,193 @@
...
```
adzialocha, line 175: Maybe we remove this commented code? Think its fine without it :+1:

adzialocha (changes requested):

aquadoggo/src/graphql/client/tests.rs
```diff
@@ -0,0 +1,210 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
```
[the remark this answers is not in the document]
cafca: lol, you previously requested _remove the newline_ and I thought you meant the one on the other side. my bad, will fix this!

aquadoggo/src/graphql/client/dynamic_query.rs
```diff
@@ -0,0 +1,556 @@
...
+            (_, None) => Err(ServerError::new(
+                "Must provide either `id` or `viewId` argument".to_string(),
+                Some(ctx.item.pos),
+            )),
+        }
+    }
+
+    /// Returns all documents for the given schema as a GraphQL value.
+    async fn query_listing(
```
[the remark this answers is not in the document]
cafca: Yes, collection is a better name! Is that from the spec? Or how is it a panda term?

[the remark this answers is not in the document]
adzialocha: Jaein, it's a term we had in the past when we had "instances" and "collections" but this somehow got forgotten. I think we say now "a collection of documents" maybe?

aquadoggo/src/graphql/client/static_types/next_entry_arguments_type.rs
cafca: I removed all the suffixes from types that are not representing some other type with the same name (e.g. `NextEntryArguments`).

For the scalars that wrap p2panda types with the exact same name I think it will be helpful to have them. I saw your approach with `scalars::X` and that also works! It's just another thing you have to keep in mind when writing code so maybe we can help panda devs in the future by adding `Scalar` or whatever GQL type is being represented as a suffix.

cafca: Hah, I have finally found the issue. The scalar types all use a default implementation of async graphql's `Scalar` trait, which doesn't validate. So I will add that instead.

aquadoggo/src/graphql/client/dynamic_query.rs
cafca: Yes, this method is called from `resolve_field` when the graphql query asks for any field that can be parsed as a schema id. The result of that parsing ends up as the `schema_id` parameter in this method. 

Now that I think about this again I realise that, in contrast to listing queries, there is no check to validate that the schema exists in the schema provider before the view for the query is retrieved, I should add that.

aquadoggo/src/db/stores/test_utils.rs
cafca: I first tried implementing this using the `Node` for p2panda but then I was missing the schema store there so that's how I ended up with this approach. I also didn't want the materialisation to happen async so that the tests don't need to wait for some magical millisecond number (that then causes failing tests when run on a very slow system).

> I like the pattern of not adding any new methods to the db store and instead modifying with helpers

Do you mean adding new methods in the sense of "adding methods to the test store that don't exist on the production store"? That seems like a good idea, yes. I mostly wanted to have a single-function interface for adding schemas and documents based on some shorthand spec. We can also move that to functions instead of methods on the store! Maybe as a follow-up?

aquadoggo/src/graphql/client/dynamic_query.rs
cafca: No, it's good you mention this. I reduced the nesting a bit.

aquadoggo/src/graphql/scalars/document_id.rs
```diff
@@ -1,29 +1,34 @@
...
+    fn try_from(document_id: DocumentId) -> Result<p2panda_rs::document::DocumentId, Self::Error> {
+        let hash = document_id.0.parse::<Hash>().map_err(Into::<Error>::into)?;
+        Ok(p2panda_rs::document::DocumentId::from(hash))
     }
 }
 
 impl Display for DocumentId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        write!(f, "{}", self.0.as_str())
```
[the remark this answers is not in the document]
cafca: First need to merge the aquadoggo validation pr for that!

adzialocha:
Ah, I see now how Scalars work 👍 Soon you know so much about async graphql that you can make a PR and integrate dynamic schemas there 🤣🤩

aquadoggo/src/db/stores/test_utils.rs
sandreae: > Do you mean adding new methods in the sense of "adding methods to the test store that don't exist on the production store"?

Yeh! This is what I meant :+1:

sandreae:
FYI: I'm halfway merging `main` into this branch, just not quite finished yet.

adzialocha (approved):
🌻🌻🌻🪐🪐🪐🌠🌠🌠

cafca:
Amazing @sandreae! I'll have a final look now.

aquadoggo/src/db/stores/test_utils.rs
cafca: I'll do a follow-up and convert the added methods to functions!

https://github.com/p2panda/aquadoggo/issues/226

sandreae:
👏👏👏👏👏👏

adzialocha:
Tadaaa!!!

cafca:
I just realized that somewhere along the way the changelog update got overrun, have to add that back.

PR #144, opened 2022-06-04
Implement `reduce` and `dependency` task logic

sandreae:
### `reduce`
- [x] fetch operations by `document_id`
- [x] fetch operations by `document_view_id`
- [x] materialise document
- [x] materialise document_views
- [x] store specific document view
- [x] store documents current view
- [x] return next dependency tasks
- [x] error handling
- [x] tests
- [x] make requested review changes

### `dependency`
- [x] ~retrieve relation graph from db (including parent and children)~ check child pinned relations are materialised
- [x] return `reduce` tasks for every missing child dependency
- [x] error on deleted documents
- [x] error handling
- [x] tests

## 📋 Checklist


aquadoggo/src/materializer/tasks/reduce.rs
```diff
@@ -1,17 +1,242 @@
...
+    let document_view_id = match &input.document_view_id {
+        Some(document_view_id) => {
+            // TODO: If we are resolving a document_view_id, but we are missing operations from it's graph,
+            // what do we want to return?
+            let document = DocumentBuilder::new(operations)
+                .build_to_view_id(Some(document_view_id.to_owned()))
+                .map_err(|_| TaskError::Critical)?;
+
+            if document.is_deleted() {
+                return Ok(None);
```
adzialocha, line 51: That's an interesting case which we haven't looked too much yet: What happens when a document gets deleted?

I haven't thought it fully through yet but I can imagine we actually want to dispatch dependency tasks for it as well as this might again have effects further down the line. For example: a deleted schema definition document will probably remove the item in the schema service (or set it to deprecated etc., we haven't decided yet).

It's good to keep it simple for now but maybe we should make a TODO / issue for that.

aquadoggo/src/materializer/tasks/reduce.rs
```diff
@@ -1,17 +1,242 @@
...
+    use p2panda_rs::storage_provider::traits::{AsStorageOperation, OperationStore};
+    use p2panda_rs::test_utils::constants::TEST_SCHEMA_ID;
+    use p2panda_rs::test_utils::fixtures::{random_document_id, random_document_view_id};
+    use rstest::rstest;
+
+    use crate::config::Configuration;
+    use crate::context::Context;
+    use crate::db::stores::test_utils::{test_db, TestSqlStore};
+    use crate::db::traits::DocumentStore;
+    use crate::materializer::tasks::reduce_task;
```
adzialocha, line 101: This you can import via `super`, then it's easy to see what objects were testing in that file 👍

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,315 @@
...
+use crate::materializer::worker::{Task, TaskError, TaskResult};
 use crate::materializer::TaskInput;
 
-pub async fn dependency_task(_context: Context, _input: TaskInput) -> TaskResult<TaskInput> {
-    Ok(None)
+pub async fn dependency_task(context: Context, input: TaskInput) -> TaskResult<TaskInput> {
+    let document_view_id = match input.document_view_id {
+        Some(id) => id,
+        // We only accept handling dependency tasks for specific document views.
+        None => return Err(TaskError::Failure),
```
adzialocha, line 11: Nice, we could make it critical as well, as it would be rather in the bug category if someone would write such code / dispatch dependency tasks without the right info.

adzialocha (commented):
Just writing some comments while we're going as I was very curious 🤩

This is really really really beautiful, the tasks are so small and powerful!

adzialocha:
The error handling is funny as we probably end up only having one critical error type or `Ok(None)` 🤣

There could be two approaches to the semantics of the return values:

- `Ok(...)` Task finished without any critical problem
- `Err(Critical)` Task failed with a critical problem

.. or:

- `Ok(...)` Task finished and did it's job
- `Err(Failure)` Task finished but couldn't do it's job
- `Err(Critical)` Task failed with a critical problem

aquadoggo/src/materializer/tasks/dependency.rs
sandreae: I actually changed the behaviour here now to handle dependency tasks for document ids as well. Not sure if we need that behaviour, but it's simple enough to get the document view and document view id we need anyway so it didn't seem like a problem. If we know we never want to dispatch a dependency task by document id we can of course bring back this error.

sandreae:
Thanks for all the comments @adzialocha! I will make those changes.

When you have time, you can also look over the dependency task logic. It seems a little more nuanced than I originally imagined. The current logic is:

- for "parent" dependencies which have been materialised (the only kind) => dispatch a `dependency` task
- for "child" dependencies have been materialised => do nothing
- for "child" dependencies have not been materialised => dispatch a `reduce` task
- for documents or document views which have all their child dependencies met => dispatch a `schema` task

adzialocha (commented):
I haven't reflected yet on the new schema store in this context but I think we currently don't need a schema task.

I'd add it as soon as we build the service 👍

sandreae:
> but I think we currently don't need a schema task with this db layout

Yep, that makes sense.

> Later we can add the schema service which will be used to query schema instances in memory (as a caching layer), it will be populated during node startup and filled with new schemas in a schema task (only if the wanted schema was missing during startup). Later we can deprecate / delete schemas in the task when the related document got deleted.

Cool, this is what I imagined we'd be doing as well :+1:

adzialocha:
> * for "parent" dependencies which have been materialised (the only kind) => dispatch a `dependency` task

This looks really close to the specification we have in the Whimsical :+1: except of this one point where I would expect this to dispatch a `reduce` task (which later on will dispatch a `dependency` task. Let's talk about it in the meeting!

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,305 @@
...
+}
+
+pub async fn dependency_task(context: Context, input: TaskInput) -> TaskResult<TaskInput> {
+    // PARSE INPUT ARGUMENTS //
+
+    // Here we retrive the document view by document id or document view id depending on what was passed into
+    // the task. This is needed so we can access the document view id and also check if a document has been
+    // deleted. We also catch invalid task inputs here (both or neither ids were passed), failing the task
+    // in a non critical way if this is the case.
+    let document_view = match (&input.document_id, &input.document_view_id) {
```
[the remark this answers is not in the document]
sandreae: Whoop! No :-) it can be refactored. I really wanted to get rid of it, but didn't think I could.

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,305 @@
...
-use crate::materializer::worker::TaskResult;
+use crate::db::traits::DocumentStore;
+use crate::materializer::worker::{Task, TaskError, TaskResult};
 use crate::materializer::TaskInput;
 
-pub async fn dependency_task(_context: Context, _input: TaskInput) -> TaskResult<TaskInput> {
-    Ok(None)
+/// helper method for retrieving a document view by it's document view id and if it doesn't exist in
+/// the store, composing a "reduce" task for this specific document view.
+async fn pinned_relation_task(
```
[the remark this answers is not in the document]
sandreae: done

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,305 @@
...
+
 use crate::context::Context;
-use crate::materializer::worker::TaskResult;
+use crate::db::traits::DocumentStore;
+use crate::materializer::worker::{Task, TaskError, TaskResult};
 use crate::materializer::TaskInput;
 
-pub async fn dependency_task(_context: Context, _input: TaskInput) -> TaskResult<TaskInput> {
-    Ok(None)
+/// helper method for retrieving a document view by it's document view id and if it doesn't exist in
```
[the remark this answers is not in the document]
sandreae: yip

cafca:
Reviewing this now

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,293 @@
...
+        vec![("one_relation_field", OperationValue::PinnedRelationList(PinnedRelationList::new([0; 2].iter().map(|_|random_document_view_id()).collect()))),
+             ("another_relation_field", OperationValue::RelationList(RelationList::new([0; 6].iter().map(|_|random_document_id()).collect())))],
+        vec![]), 2)]
+    // This document has been updated
+    #[case(test_db(4, 1, false, TEST_SCHEMA_ID.parse().unwrap(),
+        vec![("one_relation_field", OperationValue::PinnedRelationList(PinnedRelationList::new([0; 2].iter().map(|_|random_document_view_id()).collect()))),
+             ("another_relation_field", OperationValue::RelationList(RelationList::new([0; 6].iter().map(|_|random_document_id()).collect())))],
+        vec![("one_relation_field", OperationValue::PinnedRelationList(PinnedRelationList::new([0; 3].iter().map(|_|random_document_view_id()).collect()))),
+             ("another_relation_field", OperationValue::RelationList(RelationList::new([0; 10].iter().map(|_|random_document_id()).collect())))],
+    ), 3)]
```
cafca, line 136: It would be quite helpful to have some comment on each case that documents what the tested assumptions are. The `case()` syntax is very compact and terse, which can make it hard to pick apart what's going on here.

Something like

> plain relation fields should not result in dependency tasks being planned

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,293 @@
...
-use crate::materializer::worker::TaskResult;
+use crate::db::traits::DocumentStore;
+use crate::materializer::worker::{Task, TaskError, TaskResult};
 use crate::materializer::TaskInput;
 
-pub async fn dependency_task(_context: Context, _input: TaskInput) -> TaskResult<TaskInput> {
-    Ok(None)
+/// Helper method for retrieving a document view by it's document view id and if it doesn't exist in
+/// the store, composing a "reduce" task for this specific document view.
+async fn construct_relation_task(
```
cafca, line 12: You could move this function further down the file as it's a helper and instead put `dependency_task` at the top as that is what this file is about.

adzialocha (approved):

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,293 @@
...
+    // First we handle all pinned or unpinned relations defined in this tasks document view.
+    // We can think of these as "child" relations.
+    for (_key, document_view_value) in document_view.fields().iter() {
+        match document_view_value.value() {
+            p2panda_rs::operation::OperationValue::Relation(_relation) => {
+                // This is a relation to a document, if it doesn't exist in the db yet, then that
+                // means we either have no entries for this document, or we are not materialising
+                // it for some reason. We don't want to kick of a "reduce" or "dependency" task in
+                // either of these cases, however in the future we may want to flag it as a "wanted"
+                // schema.
```
cafca, line 61: Could be nice to link this to an issue

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -1,9 +1,293 @@
...
+            }
+            p2panda_rs::operation::OperationValue::RelationList(_relation_list) => {
+                // same as above...
+            }
+            p2panda_rs::operation::OperationValue::PinnedRelation(pinned_relation) => {
+                // These are pinned relations. We may have the operations for these views in the db,
+                // but this view wasn't pinned yet, so hasn't been materialised. To make sure it is
+                // materialised when possible, we dispatch a "reduce" task for any pinned relations
+                // which aren't found.
```
cafca, line 70: Not sure I understand this comment. What is the difference between the pinned and unpinned scenario that makes us act in this case and not in the other one? This could also be nice to read in a docstring for the whole function.

aquadoggo/src/materializer/tasks/reduce.rs
```diff
@@ -1,17 +1,270 @@
...
+use crate::db::traits::DocumentStore;
+use crate::materializer::worker::{Task, TaskError, TaskResult};
 use crate::materializer::TaskInput;
 
-pub async fn reduce_task(_context: Context, _input: TaskInput) -> TaskResult<TaskInput> {
-    // @TODO: Load operations from database
-    let _document = DocumentBuilder::new(vec![])
-        .build()
-        .map_err(|_| TaskError::Failure)?;
+pub async fn reduce_task(context: Context, input: TaskInput) -> TaskResult<TaskInput> {
```
cafca, line 11: Control flow in this function is quite deeply nested. So, if you want to (😉) you could factor out the logic for document id vs. document view id inputs into two separate functions to flatten it out a bit and improve code quality. However, maybe you also want to move on, and let it be as it is.

A short description of what a reduce task actually does would be nice as a docstring.

cafca (approved):
I added some documentation for the dependency task, I think the same could be very helpful for the reducer (but dinner is ready so I stop now ;) ). The control flow is quite deeply nested with multiple matches. Flattening that out would probably help make this maintainable.

Overall this is good to go though :)

aquadoggo/src/materializer/tasks/dependency.rs
adzialocha: Pinned relations point at documents which will also be represented in their latest state (from the perspective of the node). Materializing the "latest" state for a document was done in the `reduce` task (its the first thing which gets dispatched for every new incoming operation), therefore the `dependency` task doesn't need to take care of that anymore.

aquadoggo/src/materializer/tasks/dependency.rs
sandreae: I'm gunna remove the "in the future" comment, it's a little speculative right now...

sandreae:
Thanks for the review @cafca, I improved the docs and did a little flattening, I like it! Think I covered everything now.

adzialocha:
Alright! I did some crazy rebase-kung-fu and now this is ready for merging into `development`! Also, I had a go on @cafca's idea to improve the control flow nesting in `reduce_task` by factoring all into smaller methods. Please have a (hopefully) last look, especially @sandreae, I hope I didn't mess with your code too much.

Issue #156, opened 2022-06-13
`publish_entry` in GraphQL mutation does not persist operations


adzialocha:
Only the entries payload is stored but the operations table is not used.

Issue #93, opened 2022-04-29
Consider dedicated channels instead of broadcast


adzialocha:
Currently all services talk to all other services in `ServiceManager` and all worker pools get informed about all incoming tasks in `Factory`. This should not be a problem for now but later we might want to use more dedicated channels (similar to an Actor pattern) to avoid pollution.

PR #161, opened 2022-06-14
Send published entries to materializer

adzialocha:
This PR ..

* Stores operations in the database when clients publish them via the GraphQL API (#156)
* Refactors and renames all various `Context` structs as they have been used by different things on all sorts of layers, this is an attempt to clean this up, give them distinct names and separate concerns. Also, it was required to realize this PR at all!
* Sends an incoming new operation to the communication bus to inform the materializer service :sparkles:
* Makes the materializer service pick up that incoming operation and dispatch a `reducer` task :boom: 

Closes: #155, #154, #156

Depends on: #158 

## 📋 Checklist


aquadoggo/src/graphql/client/mutation.rs
```diff
@@ -29,66 +30,111 @@ impl Mutation {
...
 
-        let store = ctx.data::<SqlStorage>()?;
-        store.publish_entry(&args).await.map_err(Error::from)
+        // Validate and store entry in database
+        // @TODO: Check all validation steps here for both entries and operations. Also, there is
+        // probably overlap in what replication needs in terms of validation?
+        let response = store.publish_entry(&args).await.map_err(Error::from)?;
+
+        // Load related document from database
+        // @TODO: We probably have this instance already inside of "publish_entry"?
```
sandreae, line 48: Yes it is! It's very easy to insert the operation inside `publish_entry()` I think it requires a little thought on what error cases/behaviours we expect, but let's make that change shortly. Obvs, this is fine for now.

aquadoggo/src/graphql/client/mutation.rs
```diff
@@ -29,66 +30,111 @@ impl Mutation {
...
+        match store
+            .get_document_by_entry(&args.entry_encoded.hash())
+            .await?
+        {
+            Some(document_id) => {
+                let operation = Operation::from(&args.operation_encoded);
+                let operation_id: OperationId = args.entry_encoded.hash().into();
+
+                let verified_operation = VerifiedOperation::new(
```
sandreae, line 57: It might be neater to use `VerifiedOperation::new_from_entry(entry_encoded, operation_encoded)` here.

aquadoggo/src/graphql/client/mutation.rs
```diff
@@ -29,66 +30,111 @@ impl Mutation {
...
+                    &args.entry_encoded.author(),
+                    &operation_id,
+                    &operation,
+                )?;
+
+                // Store operation in database
+                // @TODO: This is not done by "publish_entry", maybe it needs to move there as
```
sandreae, line 64: yep, comment on this in comment above. I agree :+1:

aquadoggo/src/bus.rs
```diff
@@ -1,17 +1,15 @@
...
 /// Messages which can be sent on the communication bus.
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum ServiceMessage {
-    /// New `Entry` with an `Operation` payload arrived at the node.
-    #[allow(dead_code)]
-    NewEntryAndOperation(Entry, Operation),
+    /// A new operation arrived at the node.
+    NewOperation(OperationId),
 }
```
sandreae, line 15: Just an observation, but does the service manager never deal with entries then? Only operations?

Would there be any merit in someday having a service dedicated to storing entries and then dispatching following tasks? We might then be able to consolidate the logic for storing both entries arriving via replication and via publish entry. Haven't thought that through very far so it may not make sense.

aquadoggo/src/graphql/client/mutation.rs
```diff
@@ -29,66 +30,111 @@ impl Mutation {
...
         // Parse and validate parameters
         let args = PublishEntryRequest {
             entry_encoded: EntrySigned::new(&entry_encoded_param)?,
             operation_encoded: OperationEncoded::new(&operation_encoded_param)?,
         };
 
-        let store = ctx.data::<SqlStorage>()?;
-        store.publish_entry(&args).await.map_err(Error::from)
+        // Validate and store entry in database
+        // @TODO: Check all validation steps here for both entries and operations. Also, there is
```
sandreae, line 43: Yeh, there's an issue for this: https://github.com/p2panda/p2panda/issues/301

sandreae (approved):
Looks great, super exciting to have the final few materialisation pieces falling into place! 

I had a suggestion to change the constructor used for `VerifiedOperation`, but it's just refactoring if you'd like to. Also some comments on the TODOs and other general follow-on ideas.

I approve ;-p

aquadoggo/src/graphql/client/mutation.rs
adzialocha: Ah, thanks, didn't check if that existed, this is much better!

aquadoggo/src/bus.rs
adzialocha: Yeah, this enum is the list of all things which can be sent to all services, you're right then, with this change it will for now only be `OperationId`, but as soon as we see need for something else, we can readjust or simple just add new message types :+1:

aquadoggo/src/graphql/client/mutation.rs
adzialocha: Thats cool, though, that issue does not seem to cover validation of data but rather talks about when to throw / not throw? (I somehow have the feeling there are loads of validation cases we didn't cover yet, but maybe I'm mistaken, would be nice to have a table someday where we validate what and then just go through the code to see if its implemented [similar to how OpenMLS does it, they even have identifiers for each validation, related issue: https://github.com/p2panda/handbook/issues/126]).

aquadoggo/src/bus.rs
adzialocha: As a side note: Not all services need to process all incoming messages, so there might be different messages different services are interested in / or not. Later (piet suggested that), we can have even dedicated channels (or even an actor model) for these things, but I think thats premature optimization at this stage (https://github.com/p2panda/aquadoggo/issues/93)

PR #163, opened 2022-06-15
Fix in test db

sandreae:
:rofl: what a great mistake.

The reason it wasn't picked up is cos all validation of backlink & skiplinks happens in `publish_entry` and this is implemented (and tested) in `p2panda-rs`, so this didn't actually effect any of the tests in aquadoggo yet.

I'm going to move this test fixture across into `p2panda-rs` (which is where I caught this bug) so for now I think it's enough to just fix it here without adding tests as they will be in place when it moves.

lolz

## 📋 Checklist

- [x] ~Add this PR to the _Unreleased_ section in `CHANGELOG.md`~

### DIFF
--- a/aquadoggo/src/db/stores/test_utils.rs
+++ b/aquadoggo/src/db/stores/test_utils.rs
@@ -248,7 +248,7 @@ pub async fn test_db(
             let next_entry = Entry::new(
                 &next_entry_args.log_id,
                 Some(&next_operation),
-                next_entry_args.backlink.as_ref(),
+                next_entry_args.skiplink.as_ref(),
                 next_entry_args.backlink.as_ref(),
                 &next_entry_args.seq_num,
             )


adzialocha:
Haha! Good catch :sweat_smile: 

> The reason it wasn't picked up is cos all validation of backlink & skiplinks happens in publish_entry and this is implemented (and tested) in p2panda-rs, so this didn't actually effect any of the tests in aquadoggo yet.

It was actually tested once but got removed with the transition from JSON RPC to GraphQL, I'd love to bring it back :+1:

sandreae:
> It was actually tested once but got removed with the transition from JSON RPC to GraphQL, I'd love to bring it back +1

`publish_entry` is now implemented and all the tests for this method correctly moved over there `p2panda-rs` here: https://github.com/p2panda/p2panda/blob/main/p2panda-rs/src/storage_provider/traits/storage_provider.rs

What extra tests do you think we need here? Is it more about testing the GraphQL endpoints?

adzialocha:
> What extra tests do you think we need here? Is it more about testing the GraphQL endpoints?

It was testing a whole client - node flow of publishing entries in the same log (like 4-5 entries total). https://github.com/p2panda/aquadoggo/blob/main/aquadoggo/src/rpc/methods/publish_entry.rs#L277

sandreae:
> It was testing a whole client - node flow through the public API of publishing entries in the same log (like 4-5 entries total).

Nice, that would be good to have back :+1:

adzialocha:
I could add this as part of this PR later :+1:

adzialocha:
> > I could add this as part of this PR later +1
> 
> I think it's a different issue though, this PR is just about fixing something in the test_db setup, I think the new tests are to do with testing the graphQL implementation, I don't think they would necessarily even use the test_db.

Ah, okay! Let's merge it then.

Issue #162, opened 2022-06-15
Re-schedule unmaterialized operations during startup


adzialocha:
There is a chance that operations stay unmaterialized when the node shuts down after storing them in the database and before the worker could take care of them. For this we need to keep some state which remembers which operations have been "completed" and which are still "pending".

We could add a column for every operation in the database with a "dirty" flag. It gets set to "true" after materialization. When the node starts up it goes through the database and dispatches a `reduce` task for every operation which is still "false".

adzialocha:
Or could we find that out with our current structure? Make a query which finds operations which are not mentioned by and document view?

adzialocha:
Pinging database guru @sandreae 👨‍🏫

sandreae:
Interesting problem!

> Make a query which finds operations which are not mentioned by and document view?

We could for sure make this query, but I don't think it would solve this problem. There will be many operations in the db which aren't part of a document view as they were never materialised in a view, but they are just part of the document graph. So this query would return way too many operations, most of which we don't want to reduce again.

> We could add a column for every operation in the database with a "dirty" flag. It gets set to "true" after materialization. When the node starts up it goes through the database and dispatches a reduce task for every operation which is still "false".

This makes sense to me though! Another similar approach would be to have a dedicated table for each class of task, where operation ids would be stored. Rows are inserted when an operation is sent to the bus, and removed after the task completes. So if there are any entries in the table, we know a task didn't complete and we re-run. Could have the same pattern for dependency tasks too (i think we have the same problem there too, right?)

adzialocha:
> We could for sure make this query, but I don't think it would solve this problem. There will be many operations in the db which aren't part of a document view as they were never materialised in a view, but they are just part of the document graph. So this query would return way too many operations, most of which we don't want to reduce again.

Right, that makes totally sense, thanks!

> Another similar approach would be to have a dedicated table for each class of task, where operation ids would be stored. Rows are inserted when an operation is sent to the bus, and removed after the task completes. So if there are any entries in the table, we know a task didn't complete and we re-run.

I'm also wondering if a dedicated "system" table would be better, as I don't really see why operations need to know about the materialization state (I love that they have this "immutable" aura around them).

> Could have the same pattern for dependency tasks too (i think we have the same problem there too, right?)

[...]

PR #168, opened 2022-06-17
Reschedule tasks

adzialocha:
Closing: #162 

## 📋 Checklist


aquadoggo/src/db/stores/task.rs
```diff
@@ -0,0 +1,210 @@
...
+            ",
+        )
+        .fetch_optional(&self.pool)
+        .await
+        .map_err(|err| SqlStorageError::Transaction(err.to_string()))?;
+
+        // If yes, we are already done here
+        if task_row.is_some() {
+            return Ok(());
+        }
```
cafca, line 42: We could spare this query with a unique constraint on the tasks table couldn't we?

aquadoggo/src/db/stores/task.rs
```diff
@@ -0,0 +1,210 @@
...
+            .as_ref()
+            .map(|view_id| view_id.as_str());
+
+        // Remove task from database
+        let result = query(
+            "
+            DELETE FROM
+                tasks
+            WHERE
+                name IS $1
```
cafca, line 87: I've never seen IS used for comparison and couldn't find anything in the pg manual. Does it actually work? The standard operator is = and I only know IS as part of sth like IS DISTINCT.

aquadoggo/src/materializer/worker.rs
```diff
@@ -85,20 +85,30 @@ use std::sync::{Arc, Mutex};
...
-use tokio::sync::broadcast::{channel, Sender};
+use tokio::sync::broadcast::{channel, Receiver, Sender};
 use tokio::task;
 use triggered::{Listener, Trigger};
 
 /// A task holding a generic input value and the name of the worker which will process it
 /// eventually.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Task<IN>(pub WorkerName, IN);
```
cafca, line 95: Might as well make the worker name private now that we have a getter.

aquadoggo/src/materializer/worker.rs
```diff
@@ -506,6 +568,48 @@ mod tests {
         assert!(factory.is_empty("second"));
     }
 
+    #[tokio::test]
+    async fn on_update_subscription() {
```
cafca, line 572: This is such a nice and concise example 😌

cafca (changes requested):
Cool 😎 👍

aquadoggo/src/db/stores/task.rs
adzialocha: Yes, I know! Funnily it didn't work when I did `=` .. and using `IS` it suddenly worked. Maybe because there are `NULL` values involved?

aquadoggo/src/db/stores/task.rs
adzialocha: Yes, No :sweat_smile: When the pending tasks are loaded into the queue during rescheduling, they are not being removed from the database (yet). The worker itself will fire `on_update` with a new `TaskStatus::Pending` as soon as it got queued, which leads to a new row in the `tasks` table and ultimately to a duplicate.

The worker itself makes sure that no duplicates exist in the queues, but thats not the case with the database.

I guess one could take the rows out of the database _after_ we run  `get_tasks` and _before_ we queue them up. Or we need a new queing approach which is used for rescheduling tasks which should lead to not firing `TaskStatus::Pending`? Or of course, we just check before insertion if there is already something as it is now. :thinking:

aquadoggo/src/db/stores/task.rs
adzialocha: Indeed, this works:

```
            DELETE FROM
                tasks
            WHERE
                name = $1
                AND document_id IS $2
                AND document_view_id IS $3
```

And changing all to `=` fails as `document_id` and `document_view_id` can be `NULL` and we need to exactly match that to fullfil the query.

aquadoggo/src/materializer/worker.rs
```diff
@@ -324,6 +349,11 @@ where
         self.error_handle.clone()
     }
 
+    /// Subscribe to status changes of tasks.
+    pub fn on_update(&self) -> Receiver<TaskStatus<IN>> {
+        self.tx_status.subscribe()
+    }
+
```
sandreae, line 356: The name of this method threw me off a little, it doesn't quite describe it's function for me. I would expect something like `on_task_status_change`. I'll get used to it if there's a reason for the current naming though :+1:

sandreae (approved):
Hey hey! Great work :+1: this feels like an elegant/simple (as possible) solution to a tricky problem. Nice to have it in place already!

I had one comment on a method name but the rest is goood.

aquadoggo/src/materializer/worker.rs
adzialocha: Good idea!

aquadoggo/migrations/20220617115933_create-tasks.sql
```diff
@@ -0,0 +1,8 @@
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+CREATE TABLE IF NOT EXISTS tasks (
+    name              TEXT      NOT NULL,
+    document_id       TEXT      NULL,
+    document_view_id  TEXT      NULL,
+    PRIMARY KEY (name, document_id, document_view_id)
```
cafca, line 7: I was just stumbling about this primary key with `null` columns. If I understand correctly this is actually only possible in Sqlite:

> According to the SQL standard, PRIMARY KEY should always imply NOT NULL. Unfortunately, due to a bug in some early versions, this is not the case in SQLite. Unless the column is an INTEGER PRIMARY KEY or the table is a WITHOUT ROWID table or a STRICT table or the column is declared NOT NULL, SQLite allows NULL values in a PRIMARY KEY column. SQLite could be fixed to conform to the standard, but doing so might break legacy applications. Hence, it has been decided to merely document the fact that SQLite allows NULLs in most PRIMARY KEY columns. 

https://www.sqlite.org/lang_createtable.html

cafca (approved):

adzialocha:
Beautiful! Thats a very elegant solution, thanks @cafca for the improvements! And I'm learning something new about `SQL` again.

Issue #98, opened 2022-05-02
Correct way to handle storing operation values as `BLOB`s with `sqlx` 


sandreae:
CAVEAT: this is only if we continue to store operation values in this way, we may chose another path anyway...

Issue #31, opened 2021-05-13
CI: Tests should also run against PostgreSQL


adzialocha:
We have had a few bugs which only occur when using PostgreSQL (placeholders ala `?1` and `==` operators), it would be good to test both against SQLite but also PostgreSQL.

PR #170, opened 2022-06-22
Run tests against PostgreSQL, fix compatibility

adzialocha:
* Add CI checking tests against PostgreSQL database :green_circle: 
* Make all SQL statements compatible with SQLite and PostgreSQL
  * The queries we needed to change are commented in this PR
* Introduces `TestDatabaseRunner` which is now used in all tests to make sure the database gets disconnected after every succeeding or failing test. It changes our style of making tests slightly, but not too much:
  * Test `fn` is _not_ `async`
  * No `#[tokio::test]` attribute used, we build our own tokio runtime
  * Test is executed _within_ the new `TestDatabaseRunner::with_db_teardown` method which also gives you access to the underlying `TestDatabase` instance
* Removes MySQL support, its [complicated](https://evertpot.com/writing-sql-for-postgres-mysql-sqlite/) and we're not sure why we need it :cat2: 
* Fixes runtime of `sqlx` by switching it to `tokio` (it was still set to `async_std` and probably forgotten after the switch)
* Fixes a bunch of formatting and refactors some tests which didn't make use of the new fixtures yet

Closing: #31 and https://github.com/p2panda/aquadoggo/issues/98

## 📋 Checklist

### DIFF
--- a/aquadoggo/migrations/20220617115933_create-tasks.sql
+++ b/aquadoggo/migrations/20220617115933_create-tasks.sql
@@ -9,12 +9,11 @@ CREATE TABLE IF NOT EXISTS tasks (
 -- Create a unique index using `COALESCE`. A regular `UNIQUE` clause will
 -- consider two rows that have at least one `null` value to always be distinct
 -- but we want to check for equality including `null` values.
--- @TODO: This fails using PostgreSQL
--- CREATE UNIQUE INDEX ux_tasks ON tasks (
-    --    name,
-    --    COALESCE(document_id, 0),
-    --    COALESCE(document_view_id, 0)
-    --);
+CREATE UNIQUE INDEX ux_tasks ON tasks (
+    name,
+    COALESCE(document_id, 0),
+    COALESCE(document_view_id, 0)
+);
 
 -- Create an index because primary keys can not contain `null` columns.
 CREATE INDEX idx_tasks ON tasks (name, document_id, document_view_id);
--- a/aquadoggo/migrations/20220617115933_create-tasks.sql
+++ b/aquadoggo/migrations/20220617115933_create-tasks.sql
@@ -11,8 +11,8 @@ CREATE TABLE IF NOT EXISTS tasks (
 -- but we want to check for equality including `null` values.
 CREATE UNIQUE INDEX ux_tasks ON tasks (
     name,
-    COALESCE(document_id, 0),
-    COALESCE(document_view_id, 0)
+    COALESCE(document_id, '0'),
+    COALESCE(document_view_id, '0')
 );
 
 -- Create an index because primary keys can not contain `null` columns.
--- a/aquadoggo/migrations/20220617115933_create-tasks.sql
+++ b/aquadoggo/migrations/20220617115933_create-tasks.sql
@@ -9,11 +9,12 @@ CREATE TABLE IF NOT EXISTS tasks (
 -- Create a unique index using `COALESCE`. A regular `UNIQUE` clause will
 -- consider two rows that have at least one `null` value to always be distinct
 -- but we want to check for equality including `null` values.
-CREATE UNIQUE INDEX ux_tasks ON tasks (
-    name,
-    COALESCE(document_id, 0),
-    COALESCE(document_view_id, 0)
-);
+-- @TODO: This fails using PostgreSQL
+-- CREATE UNIQUE INDEX ux_tasks ON tasks (
+    --    name,
+    --    COALESCE(document_id, 0),
+    --    COALESCE(document_view_id, 0)
+    --);
 
 -- Create an index because primary keys can not contain `null` columns.
 CREATE INDEX idx_tasks ON tasks (name, document_id, document_view_id);
--- a/aquadoggo/src/db/mod.rs
+++ b/aquadoggo/src/db/mod.rs
@@ -21,8 +21,6 @@ pub async fn create_database(url: &str) -> Result<()> {
         Any::create_database(url).await?;
     }
 
-    Any::drop_database(url);
-
     Ok(())
 }
 
--- a/aquadoggo/src/db/models/operation.rs
+++ b/aquadoggo/src/db/models/operation.rs
@@ -23,7 +23,8 @@ pub struct OperationRow {
     /// The id of the schema this operation follows.
     pub schema_id: String,
 
-    /// The previous operations of this operation concatenated into string format with `_` seperator.
+    /// The previous operations of this operation concatenated into string format with `_`
+    /// separator.
     pub previous_operations: Option<String>,
 }
 
@@ -34,13 +35,19 @@ pub struct OperationFieldRow {
     pub operation_id: String,
 
     /// The name of this field.
-    pub name: String,
+    ///
+    /// This is an Option as a DELETE operation contains no fields.
+    pub name: Option<String>,
 
     /// The type of this field.
-    pub field_type: String,
+    ///
+    /// This is an Option as a DELETE operation contains no fields.
+    pub field_type: Option<String>,
 
     /// The actual value contained in this field.
-    pub value: String,
+    ///
+    /// This is an Option as a DELETE operation contains no fields.
+    pub value: Option<String>,
 }
 
 /// A struct representing a joined OperationRow and OperationFieldRow.
@@ -64,21 +71,22 @@ pub struct OperationFieldsJoinedRow {
     /// The id of the schema this operation follows.
     pub schema_id: String,
 
-    /// The previous operations of this operation concatenated into string format with `_` seperator.
+    /// The previous operations of this operation concatenated into string format with `_`
+    /// separator.
     pub previous_operations: Option<String>,
 
     /// The name of this field.
     ///
-    /// This is an Option as a delete operation contains no fields.
+    /// This is an Option as a DELETE operation contains no fields.
     pub name: Option<String>,
 
     /// The type of this field.
     ///
-    /// This is an Option as a delete operation contains no fields.
+    /// This is an Option as a DELETE operation contains no fields.
     pub field_type: Option<String>,
 
     /// The actual value contained in this field.
     ///
-    /// This is an Option as a delete operation contains no fields.
+    /// This is an Option as a DELETE operation contains no fields.
     pub value: Option<String>,
 }
--- a/aquadoggo/src/db/models/operation.rs
+++ b/aquadoggo/src/db/models/operation.rs
@@ -68,11 +68,17 @@ pub struct OperationFieldsJoinedRow {
     pub previous_operations: Option<String>,
 
     /// The name of this field.
-    pub name: String,
+    ///
+    /// This is an Option as a delete operation contains no fields.
+    pub name: Option<String>,
 
     /// The type of this field.
-    pub field_type: String,
+    ///
+    /// This is an Option as a delete operation contains no fields.
+    pub field_type: Option<String>,
 
     /// The actual value contained in this field.
-    pub value: String,
+    ///
+    /// This is an Option as a delete operation contains no fields.
+    pub value: Option<String>,
 }
--- a/aquadoggo/src/db/models/operation.rs
+++ b/aquadoggo/src/db/models/operation.rs
@@ -24,7 +24,7 @@ pub struct OperationRow {
     pub schema_id: String,
 
     /// The previous operations of this operation concatenated into string format with `_` seperator.
-    pub previous_operations: String,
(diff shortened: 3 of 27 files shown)


aquadoggo/src/db/mod.rs
```diff
@@ -21,8 +21,6 @@ pub async fn create_database(url: &str) -> Result<()> {
         Any::create_database(url).await?;
     }
 
-    Any::drop_database(url);
```
adzialocha, line 24: What was that? :scream:

aquadoggo/src/db/stores/document.rs
```diff
@@ -438,7 +441,10 @@ mod tests {
             .await
             .unwrap();
 
-        assert!(view_does_not_exist.is_none())
+        assert!(view_does_not_exist.is_none());
+
+        // Disconnect from database
+        db.close().await;
```
adzialocha, line 447: We need to add this to every test, to make sure the connection is dropped before we can move over to the next test.

aquadoggo/src/db/stores/operation.rs
```diff
@@ -144,7 +144,7 @@ impl OperationStore<VerifiedOperation> for SqlStorage {
                             .bind(name.to_owned())
                             .bind(value.field_type().to_string())
                             .bind(db_value)
-                            .bind(index.to_string())
+                            .bind(index as i32)
```
adzialocha, line 147: :middle_finger: PostgreSQL was complaining that this column is actually a number (`NUMERIC`).

aquadoggo/migrations/20220617115933_create-tasks.sql
```diff
@@ -9,11 +9,12 @@ CREATE TABLE IF NOT EXISTS tasks (
 -- Create a unique index using `COALESCE`. A regular `UNIQUE` clause will
 -- consider two rows that have at least one `null` value to always be distinct
 -- but we want to check for equality including `null` values.
-CREATE UNIQUE INDEX ux_tasks ON tasks (
-    name,
-    COALESCE(document_id, 0),
-    COALESCE(document_view_id, 0)
-);
+-- @TODO: This fails using PostgreSQL
+-- CREATE UNIQUE INDEX ux_tasks ON tasks (
```
adzialocha, line 13: :point_up_2: Upsi, this doesn't seem to work on PostgreSQL :cry:

adzialocha: Since the migrations fail all tests fail currently

sandreae:
Thanks for doing all this!!

> whenever a test panics it leaves the connection to the database alive

Do you know if this also happens when we are testing for expected panics?

Like here for example: https://github.com/p2panda/aquadoggo/blob/5a1a845ab1b6af8dcdb5ac60b94f193dbb614ad4/aquadoggo/src/db/stores/schema.rs#L169

> I couldn't find a way to disconnect from databases automatically when something fails, except of manually catching every error and doing it manually .. maybe you have an idea?

Hmmm, I had a little look, couldn't find much in the way of "teardown" support for tests. There's this: https://docs.rs/test-context/latest/test_context/ but I believe it clashes with `rstest`.....

> Where apparently previous_operations is a NULL column but in the deserialized OperationRow its not an Option. Funny, that SQLite didn't cry about this sweat_smile

Weird!

> .. and wow, what was that? https://github.com/p2panda/aquadoggo/pull/170#discussion_r904898722 cold_sweat

Weirder!!! Actually, I don't get what's going on there. We were dropping the db straight away..?

adzialocha:
> Do you know if this also happens when we are testing for expected panics?

I haven't come there yet, good question! I assume its alright as it will lead to an "ok" test.

>  I had a little look, couldn't find much in the way of "teardown" support for tests

This also looks interesting, very simple and not as cool as an attribute but maybe compatible with `rtest`: https://medium.com/all-about-rust/example-for-setup-and-teardown-in-testing-using-rust-3606bca60a31

sandreae:
> This also looks interesting, very simple and not as cool as an attribute but maybe compatible with rtest: https://medium.com/all-about-rust/example-for-setup-and-teardown-in-testing-using-rust-3606bca60a31

Well that looks very smart and simple :+1:

sandreae:
Identified another failure: testing for exact error strings returned from db fails because different databases return different errors...

adzialocha:
And now I've had a go on automatically unwinding every test, which seems to work .. Sadly there are some issues around fixtures which return stack-allocated data (like `&str`) and the new test runner requires them to implement `Clone`, I suggest we change them to `String`?

sandreae:
Woah, this looks good. I was working on it too but not with much success. This solution is nice.

sandreae:
We still have loooots of errors, but now we can see what they are :-)

sandreae:
> I'm not sure if it works with the should panic attributes though .. Also, I haven't updated yet the tests which do not make use of the new TestDatabaseRunner.

Hmm, yeh, it doesn't work with this.

sandreae:
> > I'm not sure if it works with the should panic attributes though .. Also, I haven't updated yet the tests which do not make use of the new TestDatabaseRunner.
> 
> Hmm, yeh, it doesn't work with this.

I'm just gunna get rid of them all.

aquadoggo/src/graphql/client/mutation.rs
```diff
@@ -407,84 +414,176 @@ mod tests {
...
+                    .unwrap()
+                    .public_key()
+                    .to_owned(),
             )
-        }
+            .unwrap();
+
+            // Setup the server and client with a new empty store.
+            let (tx, _rx) = broadcast::channel(16);
+            let new_store = initialize_store().await;
```
sandreae, line 474: This test uses 2 stores which causes problems with multiple db connections. I imagine it can just be refactored somehow.

aquadoggo/src/db/models/operation.rs
```diff
@@ -24,7 +24,7 @@ pub struct OperationRow {
     pub schema_id: String,
 
     /// The previous operations of this operation concatenated into string format with `_` seperator.
-    pub previous_operations: String,
+    pub previous_operations: Option<String>,
```
sandreae, line 27: Now an option

aquadoggo/src/db/models/operation.rs
```diff
@@ -65,14 +65,20 @@ pub struct OperationFieldsJoinedRow {
...
 
     /// The name of this field.
-    pub name: String,
+    ///
+    /// This is an Option as a delete operation contains no fields.
+    pub name: Option<String>,
```
sandreae, line 73: Also `name` `field_type` and `value` are options (to account for delete operations)

aquadoggo/src/db/stores/entry.rs
```diff
@@ -348,7 +345,7 @@ impl EntryStore<StorageEntry> for SqlStorage {
             WHERE
                 author = $1
                 AND log_id = $2
-                AND CAST(seq_num AS NUMERIC) BETWEEN $3 and $4
+                AND CAST(seq_num AS NUMERIC) BETWEEN CAST($3 AS NUMERIC) and CAST($4 AS NUMERIC)
```
sandreae, line 348: Makes the range comparison work

aquadoggo/src/db/stores/task.rs
```diff
@@ -62,8 +63,8 @@ impl SqlStorage {
             WHERE
                 name = $1
                 -- Use `IS` because these columns can contain `null` values.
-                AND document_id IS $2
-                AND document_view_id IS $3
+                AND COALESCE(document_id, '') = COALESCE($2, '')
```
sandreae, line 66: `IS` doesn't seem to be supported in postgres so this is a workaround to use `=`

sandreae:
I did refactoring of the tests which din't use the runner yet, and some postgres compatibility fixes.

adzialocha:
:open_mouth:

aquadoggo/src/db/stores/task.rs
adzialocha: For some reason it did not allow me to use an empty string?

aquadoggo/migrations/20220617115933_create-tasks.sql
adzialocha: Fixed now via:

```sql
CREATE UNIQUE INDEX ux_tasks ON tasks (
    name,
    COALESCE(document_id, '0'),
    COALESCE(document_view_id, '0')
);
```

aquadoggo/src/db/stores/test_utils.rs
```diff
@@ -248,7 +248,7 @@ impl TestDatabaseRunner {
             pool.close().await;
 
             // Panic here when test failed to propagate it further
-            assert!(result.is_ok());
+            result.unwrap();
```
adzialocha, line 251: I believe this will make `#[should_panic]` work again

sandreae: Ah, DOH!!! Yeh, hopefully, why didn't I think of that...

aquadoggo/migrations/20220617115933_create-tasks.sql
```diff
@@ -11,8 +11,8 @@ CREATE TABLE IF NOT EXISTS tasks (
 -- but we want to check for equality including `null` values.
 CREATE UNIQUE INDEX ux_tasks ON tasks (
     name,
-    COALESCE(document_id, 0),
-    COALESCE(document_view_id, 0)
+    COALESCE(document_id, '0'),
+    COALESCE(document_view_id, '0')
```
adzialocha, line 15: This needs to be a string

aquadoggo/src/db/stores/entry.rs
```diff
@@ -435,259 +431,247 @@ mod tests {
...
+            )
+            .unwrap();
 
-        assert_eq!(
-            result.unwrap_err().to_string(),
-            "Error occured during `EntryStorage` request in storage provider: error returned from \
-            database: UNIQUE constraint failed: entries.author, entries.log_id, entries.seq_num"
-        )
+            let result = db.store.insert_entry(duplicate_doggo_entry).await;
+            assert!(result.is_err());
```
adzialocha, line 472: Returned error strings are different depending on the used database, so we need to just check via `is_err`

aquadoggo/src/db/stores/test_utils.rs
adzialocha: I've tested it, it works again :+1:

aquadoggo/src/db/mod.rs
adzialocha: I think it was never `await`-ed, therefore it also never worked, but 😱

sandreae (approved):
I'm really happy with this now. 

Regarding the point I mentioned on the call earlier, I think seeing as we really are use `rstest` everywhere then there's no need to add another option now.

aquadoggo/src/db/stores/test_utils.rs
```diff
@@ -168,23 +173,97 @@ pub async fn insert_entry_operation_and_view(
...
+
+    /// The fields used for every CREATE operation.
+    create_operation_fields: Vec<(&'static str, OperationValue)>,
+
+    /// The fields used for every UPDATE operation.
+    update_operation_fields: Vec<(&'static str, OperationValue)>,
+}
+
+impl TestDatabaseRunner {
+    pub fn with_db_teardown<F: AsyncTestFn + Send + Sync + 'static>(&self, test: F) {
```
cafca, line 213: A docstring here could be friendly for when you encounter this function in some test.

cafca (approved):
I had two wishes for docstrings and added some myself, otherwise it's all good!

aquadoggo/src/db/stores/test_utils.rs
```diff
@@ -168,23 +173,97 @@ pub async fn insert_entry_operation_and_view(
...
+            });
+
+            // Get a handle of the task so we can use it later
+            let result = handle.await;
+
+            // Unwind the test by closing down the connection to the database pool. This will
+            // be reached even when the test panicked
+            pool.close().await;
+
+            // Panic here when test failed to propagate it further
```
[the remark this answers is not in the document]
adzialocha: Not sure if I understand your question, but I'll try to explain it from another angle: Panicks in tasks are not propagated further (the word here is "unwinded"), they just crash the task. To allow the test runner to understand that something happend we have to take the task handle, unwrap it to inform the "parent" about its failure.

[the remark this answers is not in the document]
adzialocha: The function of this is to make the test fail when it should :-D

PR #176, opened 2022-06-28
Refactor test db creation

sandreae:
Small bit of refactoring:

- Introduced `send_to_store()` method which helps in publishing operations encoded on entries to the store

## 📋 Checklist

- [x] Add tests that cover your changes
- [x] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [x] Link this PR to any issues it closes
- [x] New files contain a SPDX license header

PR #177, opened 2022-06-28
Fix bug when materialising documents containing (non-existent) pinned relations

sandreae:
When the materialiser comes across a pinned relations it kicks off a reduce task for each document view being related to. This is the correct behaviour. But then our clever reduce task was freaking out if there was no document for the requested document view already in the store. This is buggy behaviour. This PR fixes it by not critically failing the task but instead exiting the task silently. Also adds a test for this case.

branch includes unmerged changes from: https://github.com/p2panda/aquadoggo/pull/175 + https://github.com/p2panda/aquadoggo/pull/176

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/materializer/tasks/reduce.rs
+++ b/aquadoggo/src/materializer/tasks/reduce.rs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
 use log::debug;
-
 use p2panda_rs::document::{DocumentBuilder, DocumentId, DocumentViewId};
 use p2panda_rs::operation::VerifiedOperation;
 use p2panda_rs::storage_provider::traits::OperationStore;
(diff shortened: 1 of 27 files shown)


aquadoggo/src/materializer/tasks/reduce.rs
```diff
@@ -17,15 +19,26 @@ use crate::materializer::TaskInput;
...
+
     // Find out which document we are handling
-    let document_id = resolve_document_id(&context, &input).await?;
+    let document_id = match resolve_document_id(&context, &input).await? {
+        Some(document_id) => Ok(document_id),
+        None => {
+            debug!("No document found for this view, exit without dispatching any other tasks");
+            return Ok(None);
+        }
+    }?;
```
sandreae, line 31: Instead of returning `Ok(None)` when a document for a requested view didn't exist, we were critically failing the task here.

Logic bug....

sandreae:
I think the test coverage drops here because all the `debug!()` lines add more untested lines when logging after a critical db error is found. We don't test these cases right now as they're hard to simulate. We could use mocking for this I recon.

adzialocha (approved):

Issue #182, opened 2022-06-30
Remove `mockall`


adzialocha:
Its only used in one test and we can mock quite well with `rstest` and our traits

PR #137, opened 2022-05-28
Graphql replication client

pietgeursen:
Describe your changes here.

## 📋 Checklist

- [ ] Add tests that cover your changes
- [ ] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [ ] Link this PR to any issues it closes
- [ ] New files contain a SPDX license header

Issue #120, opened 2022-05-17
Remove unused dependencies


cafca:
https://github.com/p2panda/aquadoggo/pull/101#discussion_r867754194

PR #184, opened 2022-06-30
Refactor replication API

adzialocha:
This PR refactors the Replication API:

* Removes concept of aliased authors: Its a good idea but at this point a too complex optimization. Also it probably needs a moment in the boiling hot pot of the p2panda specification process :cat2: 
* Moves `EntryArgsRequest` and `PublishEntryRequest` into `db` module as it is more a struct used to query the database than an actual GraphQL thing
* Consistent access of `Context`: `ReplicationContext` has been removed in favor of accessing already existing `SqlStorage`
* Replaces `graphql-client` with simpler `gql-client`, removes need of external `.graphql` files, `bin` generator and auto-generated types
* Removes last test using `automock` by replacing it with `rstest` mocking approach
* Updates GraphQL field and params names according to the latest specification
* Moves scalars out of `replication` as they should be used both by `client` and `replication`
* Serde methods to represent `LogId` and `SeqNum` as strings are moved directly next to the regarding scalar implementations
* Added some docstrings and GraphQL descriptions
* Added some more tests
* Updated dependencies and removed unused ones

Depends on #137 

Closing #180 and #182 and #120

## 📋 Checklist


aquadoggo/src/graphql/replication/query.rs
```diff
@@ -0,0 +1,421 @@
...
+                        &log_id.into(),
+                        &start_seq_num,
+                        max_number_of_entries,
+                    )
+                    .await?
+                    .into_iter()
+                    .map(|entry_and_operation| {
+                        // Unwrap as the entry was already validated before it entered the database
+                        //
+                        // @TODO: Is there a more efficient way than decoding the entry to access
```
sandreae, line 120: Yes! You can use `entry_and_operation.seq_num()` ;-p

Right now, that actually just decodes the entry under-the-hood... but that is because we have a rather inefficient implementation of `AsStorageEntry` which we can improve later with a small amount of effort.

aquadoggo/src/graphql/replication/query.rs
```diff
@@ -0,0 +1,421 @@
...
+            // author
+            let public_key: Author = db
+                .test_data
+                .key_pairs
+                .first()
+                .unwrap()
+                .public_key()
+                .to_owned()
+                .try_into()
+                .unwrap();
```
sandreae, line 264: This is so ugly... :rofl: I think we need some more conversion traits for `Author`.

sandreae (approved):
This looks great to me, nice refactoring :star2: 

I like `entry_encoded` => `entry` and `operation_encoded` => `operation` a LOT!

I had a couple of comments, one of which I have fix with a commit. I'll raise issues for the other two.

PR #192, opened 2022-07-05
Don't error in SchemaStore when schema fields are missing

sandreae:
Currently we return an error from `get_schema` or `get_all_schema` in `SchemaStore` if a schema cannot be constructed due to it's fields not being present in the database. This isn't the behaviour we expect, rather we want to ignore any schemas which can't be fully constructed.

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/db/stores/schema.rs
+++ b/aquadoggo/src/db/stores/schema.rs
@@ -39,20 +39,15 @@ impl SchemaStore for SqlStorage {
             let scheme_field_view: SchemaFieldView =
                 match self.get_document_view_by_id(&field_id).await? {
                     Some(document_view) => document_view.try_into()?,
-                    None => {
-                        return Err(SchemaStoreError::MissingSchemaFieldDefinition(
-                            field_id,
-                            id.to_owned(),
-                        ))
-                    }
+                    None => return Ok(None),
                 };
 
             schema_fields.push(scheme_field_view);
         }
 
-        let schema = Schema::from_views(schema_view, schema_fields)?;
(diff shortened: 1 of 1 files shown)


aquadoggo/src/db/stores/schema.rs
```diff
@@ -331,21 +322,18 @@ mod tests {
...
+                &db.store,
+                &key_pair,
+                &schema_fields_id,
+                schema_definition,
+            )
+            .await;
 
             // Retrieve the schema by it's document_view_id.
-            let schema = db.store.get_schema_by_id(&document_view_id).await;
+            let schema = db.store.get_schema_by_id(&document_view_id).await.unwrap();
```
adzialocha, line 334: Maybe add comment why you `unwrap` here

aquadoggo/src/db/stores/schema.rs
```diff
@@ -39,20 +39,15 @@ impl SchemaStore for SqlStorage {
...
-                        ))
-                    }
+                    None => return Ok(None),
                 };
 
             schema_fields.push(scheme_field_view);
         }
 
-        let schema = Schema::from_views(schema_view, schema_fields)?;
+        let schema = Schema::from_views(schema_view, schema_fields).ok();
```
adzialocha, line 48: Maybe add comment why you "silently fail" / convert to an `Option` here.

adzialocha (changes requested):
Nice! Thank you for fixing this!

I'd add some more comments on why certain things are the way there are, but otherwise its ready to go!

adzialocha (approved):

PR #218, opened 2022-08-01
Add service ready signal

cafca:
Services receive an additional parameter with a oneshot channel to which they send a unit value once they are ready to process messages on the communication bus. This allows us to understand when a service becomes ready from outside the service. Previously there were a couple of tests that waited for some arbitrary time to allow services to become available.

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/http/service.rs
+++ b/aquadoggo/src/http/service.rs
@@ -8,14 +8,13 @@ use axum::http::Method;
 use axum::routing::get;
 use axum::Router;
 use log::{debug, warn};
-use tokio::sync::oneshot;
 use tower_http::cors::{Any, CorsLayer};
 
 use crate::bus::ServiceSender;
 use crate::context::Context;
 use crate::http::api::{handle_graphql_playground, handle_graphql_query};
 use crate::http::context::HttpServiceContext;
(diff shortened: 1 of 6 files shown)


aquadoggo/src/http/service.rs
```diff
@@ -38,7 +40,12 @@ pub fn build_server(http_context: HttpServiceContext) -> Router {
 }
 
 /// Start HTTP server.
-pub async fn http_service(context: Context, signal: Shutdown, tx: ServiceSender) -> Result<()> {
+pub async fn http_service(
+    context: Context,
+    signal: Shutdown,
+    tx: ServiceSender,
+    tx_ready: oneshot::Sender<()>,
```
adzialocha, line 47: Maybe give the sender also an own type like you already do for the receiver (`ServiceReady`)? I think that would also ready easier as well.

adzialocha (changes requested):
Just requested some small changes, otherwise this look super :+1:

aquadoggo/src/http/service.rs
cafca: Like this?

adzialocha (approved):

PR #230, opened 2022-08-11
The monster refactor

sandreae:
Integrate new API for the "monster" https://github.com/p2panda/p2panda/pull/415

- new validation methods
  - introduce validating operations against schema 
  - rewrite all validation in GraphQL / `publish()` / `next_args()`
  - replace bamboo verify with new validation methods
- `StorageEntry`
  - no longer contains it's payload `EncodedOperation`, need to refactor in many places where this is expected
  - rework constructor so it can't fail 
- `ValidatedOperation`
  - we can no longer directly construct this so we need an own type in `aquadoggo` representing operations coming from the db
- `Operation`
  - constructors now accept fields as vec (introduce conversion trait for `OperatoinFields`?)
  - `::new_create()`, `::new_update()`, `::new_delete()` no longer exist, use `OperationBuilder()` instead
- `EntryEncoded` -> `Operation` decoding difficult now but we rely on this in some testing patterns
- `EncodedEntryAndOperation` replication needs operation from store

## 📋 Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -538,24 +538,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2fff2a6927b3bb87f9595d67196a70493f627687a71d87a0d692242c33f58c11"
 
 [[package]]
-name = "cddl-cat"
-version = "0.6.1"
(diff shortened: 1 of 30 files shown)


cafca:
🦫

PR #231, opened 2022-08-11
Friendly names for scalars in GraphQL schema

### LINKED ISSUE #229: GraphQL Docs look funny with `Scalar` suffixes
Screenshot 2022-08-11 at 12-01-58 Playground - http __localhost 2020_graphql

Is there a way to tell `async-graphql` that we want a different name for the scalar? It looks like it currently derives it from the struct names.

cafca:
Removes `Scalar` suffix in gql schema.

Closes #229

## 📋 Checklist


cafca:
@adzialocha the other odd one is `EntryHash` as we just call that `Hash` everywhere else. We could also rename it for consistency but on the other hand having a type just called "hash" might also be confusing. The type's description tells you what it is but still. What do you think?

adzialocha (approved):
> @adzialocha the other odd one is `EntryHash` as we just call that `Hash` everywhere else. We could also rename it for consistency but on the other hand having a type just called "hash" might also be confusing. The type's description tells you what it is but still. What do you think?

I agree, let's stick to `EntryHash`. Its also the only "Hash" we provide right now via the API.

PR #235, opened 2022-08-12
Untagged operation format & schema validation

sandreae:
Integrate new API for the "monster" https://github.com/p2panda/p2panda/pull/415 including some changes needed for this PR here https://github.com/p2panda/p2panda/pull/434

- [x] implement all simple API changes implemented here: https://github.com/p2panda/p2panda/pull/415
- [x] refactor tests which are simple to update
- [x] remove tests which are hard to update
- [x] update `domain` methods to use new validation methods
- [x] fetch schema from provider before passing it, entry and operation, into `publish`
- [x] correct method for parsing `EncodedEntry` + `EncodedOperation` in gql scalar traits
- [x] re-implement storing entry + payloads which arrive via replication
- [x] re-implement test utilities `send_to_store` etc...
- [x] fix existing tests which are failing
- [x] bring back all deleted tests
- [x] add tests for schema validation
- [x] update doc strings for `publish`
- [x] update `nextEntryArgs` to accept a document view id
- [x] e2e test poetry :pancakes: 

# Next up
- there is still this mysterious GraphQL issue: https://github.com/p2panda/aquadoggo/issues/242

## 📋 Checklist


aquadoggo/src/graphql/replication/response.rs
```diff
@@ -97,15 +116,7 @@ mod tests {
...
-            let author: Author = db
-                .test_data
-                .key_pairs
-                .first()
-                .unwrap()
-                .public_key()
-                .to_owned()
-                .try_into()
-                .unwrap();
+            let author: Author = db.test_data.key_pairs.first().unwrap().public_key().into();
```
adzialocha, line 119: Haha, fantastic 🤣

aquadoggo/src/graphql/scalars/document_id_scalar.rs
```diff
@@ -4,10 +4,9 @@ use std::fmt::Display;
 
 use async_graphql::{InputValueError, InputValueResult, Scalar, ScalarType, Value};
 use p2panda_rs::document::DocumentId;
-use serde::{Deserialize, Serialize};
 
 /// Id of a p2panda document.
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
adzialocha, line 9: Nice, I think we can remove them from all scalars 👍 I'll make a ticket.

aquadoggo/src/graphql/client/mutation.rs
```diff
@@ -338,128 +299,128 @@ mod tests {
...
-        "Failed to parse \"EntrySignedScalar\": invalid hex encoding in entry"
+        "Failed to parse \"EntrySignedScalar\": Invalid character '-' at position 0"
     )]
     #[case::no_operation(
-        &ENTRY_ENCODED,
-        "",
-        "operation needs to match payload hash of encoded entry"
+        &EncodedEntry::from_bytes(&ENTRY_ENCODED).to_string(),
+        "".as_bytes(),
+        "cbor decoder failed failed to fill whole buffer"
```
sandreae, line 318: Is this the error we expect for this case? It seems to come from a cbor IO error: https://github.com/p2panda/p2panda/blob/3ea76a53bac01a073c226cd4b64085a58bf14ae2/p2panda-rs/src/operation/decode.rs#L35

aquadoggo/src/domain.rs
```diff
@@ -202,44 +208,60 @@ pub async fn next_args<S: StorageProvider>(
...
+    // VALIDATE ENTRY AND OPERATION //
+    //////////////////////////////////
+
+    // @TODO: Check this validation flow is still correct.
 
     // Verify that the claimed seq num matches the expected seq num for this author and log.
     let latest_entry = store.get_latest_entry(&author, log_id).await?;
     let latest_seq_num = latest_entry.as_ref().map(|entry| entry.seq_num());
-    is_next_seq_num(latest_seq_num.as_ref(), seq_num)?;
+    is_next_seq_num(latest_seq_num, seq_num)?;
```
sandreae, line 234: This step is not needed I don't think as `validate_operation_with_entry()` will catch the case where the encoded seq num doesn't match the expected one.

aquadoggo/src/domain.rs
```diff
@@ -202,44 +208,60 @@ pub async fn next_args<S: StorageProvider>(
...
     // The backlink for this entry is the latest entry from this public key's log.
     let backlink = latest_entry;
 
     // If a skiplink is claimed, get the expected skiplink from the database, errors
     // if it can't be found.
-    let skiplink = match entry.skiplink_hash() {
+    let skiplink = match entry.skiplink() {
         Some(_) => Some(get_expected_skiplink(store, &author, log_id, seq_num).await?),
         None => None,
     };
```
sandreae, line 244: This flow seems to not match now. What is currently happening is that we fetch the `expected` skiplink based on the author and seq num encoded on the passed entry. Would we rather just fetch the claimed skiplink based on the encoded skiplink hash?

sandreae (commented):
Just doing a review of `publish` specifically looking at how the new validation flow effects what was already there.

sandreae:
clippy is so unhappy, lolz

aquadoggo/src/domain.rs
adzialocha: Why does it not match?

aquadoggo/src/graphql/client/mutation.rs
adzialocha: Oh yeah, the error makes sense but the message is not so nice. We shouldn't change the IO error message but we can check if the array is empty before it arrives there.

aquadoggo/src/domain.rs
sandreae: Well, as I understand it, in [`validate_log_integrity()`](https://github.com/p2panda/p2panda/blob/53e74101c00022e46c35093a78f049431b41376d/p2panda-rs/src/entry/validate.rs#L48) we are expecting the claimed skiplink rather than the expected.

sandreae: Either way the validation works, but maybe the flow would be in a different order depending on the two approaches?

sandreae:
I added a lot more comments and did some refactoring in the store tests. Want to do more elsewhere, will see if I get round to it. But please, during review, just mention any specific places where we should add more comments :+1:

aquadoggo/src/db/provider.rs
```diff
@@ -105,66 +97,13 @@ impl SqlStorage {
...
-            let result = db
-                .store
-                .get_schema_by_document_view(&document_view_id)
-                .await;
-
-            assert!(result.is_ok());
-            assert_eq!(result.unwrap().unwrap().name(), "venue");
-        });
-    }
+    // @TODO: bring back test_get_schema_for_view test
```
sandreae, line 106: Bring back test

aquadoggo/src/graphql/client/dynamic_types/tests.rs
```diff
@@ -134,88 +134,88 @@ fn application_schema_container_type(#[from(test_db)] runner: TestDatabaseRunner
...
-                        "name": "schema_definition_v1"
-                    }
-                }]
-            }
-        });
-
-        assert_eq!(response.data, expected_data, "\n{:#?}\n", response.errors);
-    });
-}
+// @TODO: Failing test
```
sandreae, line 137: Bring back test

aquadoggo/src/materializer/service.rs
```diff
@@ -134,384 +134,4 @@ pub async fn materializer_service(
...
-                .store
-                .get_document_by_id(&entry_encoded.hash().into())
-                .await
-                .unwrap()
-                .expect("We expect that the document is `Some`");
-            assert_eq!(document.id(), &entry_encoded.hash().into());
-        });
-    }
-}
+// @TODO: Bring back all service tests
```
sandreae, line 137: Bring back tests

sandreae:
Okay, all tests are back, but one is failing for reasons unknown.... Do you mind just looking at that @adzialocha ?

Once that is done I would say it is ready for review :fireworks:

aquadoggo/src/db/provider.rs
```diff
@@ -33,13 +29,9 @@ impl SqlStorage {
 /// databases.
 #[async_trait]
 impl StorageProvider for SqlStorage {
-    type EntryArgsRequest = EntryArgsRequest;
-    type EntryArgsResponse = EntryArgsResponse;
-    type PublishEntryRequest = PublishEntryRequest;
-    type PublishEntryResponse = PublishEntryResponse;
     type StorageLog = StorageLog;
```
sandreae, line 32: Rename to Log

aquadoggo/src/db/stores/document.rs
```diff
@@ -436,73 +403,19 @@ mod tests {
         runner: TestDatabaseRunner,
     ) {
         runner.with_db_teardown(|db: TestDatabase| async move {
+            // We try to retrieve a document view by it's id but now view
```
sandreae, line 406: Typo, "now" should be "no".

aquadoggo/src/db/stores/operation.rs
```diff
@@ -4,33 +4,99 @@ use std::collections::BTreeMap;
...
+    }
+}
+
+impl AsOperation for StorageOperation {
+    /// Returns action type of operation.
+    fn action(&self) -> OperationAction {
+        self.action.to_owned()
+    }
+
+    /// Returns schema if of operation.
```
sandreae, line 63: Typo

aquadoggo/src/db/utils.rs
```diff
@@ -21,71 +24,64 @@ use crate::db::models::OperationFieldsJoinedRow;
 /// when retrieving an operation from the db.
 pub fn parse_operation_rows(
```
sandreae, line 25: Even though these parsing methods expect valid values to be coming from the dB, it would be nicer to catch and return errors, then unwrap the result when calling this method.

aquadoggo/src/e2e_tests.rs
```diff
@@ -0,0 +1,402 @@
...
+use p2panda_rs::operation::traits::Actionable;
+use p2panda_rs::operation::{Operation, OperationAction, OperationBuilder, OperationValue};
+use p2panda_rs::schema::{FieldType, Schema, SchemaId};
+use reqwest::Client;
+use serde_json::{json, Map, Value};
+
+use crate::{Configuration, Node};
+
+#[tokio::test]
+async fn e2e() {
```
sandreae, line 24: Could do with some more emojis 🤣🤣🤣

sandreae (commented):
Just a few typos and follow up comments.

sandreae:
This is the test which fails: https://github.com/p2panda/aquadoggo/blob/80e217163dd5021789788be0a2a6a651950837c2/aquadoggo/src/graphql/client/dynamic_types/tests.rs#L74

With this error:

```
thread 'graphql::client::dynamic_types::tests::application_schema_fields_type' panicked at 'Publish schema fields: ValidateOperationError(SchemaValidation(InvalidSchemaFieldDefinition(TypeInvalid)))', aquadoggo/src/db/stores/test_utils/store.rs:110:14
```

aquadoggo/src/graphql/client/static_query.rs
```diff
@@ -27,30 +26,26 @@ impl StaticQuery {
...
+            desc = "Document the entry's UPDATE or DELETE operation is referring to, \
+            can be left empty when it is a CREATE operation"
+        )]
+        document_view_id: Option<scalars::DocumentViewIdScalar>,
+        // @TODO: Figure out why this fixes things....
         #[graphql(
             name = "documentId",
             desc = "Document the entry's UPDATE or DELETE operation is referring to, \
             can be left empty when it is a CREATE operation"
         )]
```
sandreae, line 40: Taking out `documentId` here breaks something inside the graphql magic, even though we are already using `documentViewId` in all cases.

sandreae: See issue: https://github.com/p2panda/aquadoggo/issues/242

adzialocha:
> This is the test which fails:
> 
> https://github.com/p2panda/aquadoggo/blob/80e217163dd5021789788be0a2a6a651950837c2/aquadoggo/src/graphql/client/dynamic_types/tests.rs#L74
> 
> With this error:
> 
> ```
> thread 'graphql::client::dynamic_types::tests::application_schema_fields_type' panicked at 'Publish schema fields: ValidateOperationError(SchemaValidation(InvalidSchemaFieldDefinition(TypeInvalid)))', aquadoggo/src/db/stores/test_utils/store.rs:110:14
> ```

I think I found the problem and fixed it here: https://github.com/p2panda/p2panda/issues/452

adzialocha:
I receive random failing tests when I run `cargo test` for the first times (very much looking like race conditions):

1.

```
---- db::stores::entry::tests::paginated_log_entries::case_2 stdout ----
-------------- TEST START --------------
thread 'db::stores::entry::tests::paginated_log_entries::case_2' panicked at 'Send to store: Custom("Entry's claimed seq num of 1 does not match expected seq num of 4 for given author and log")', /home/adz/.cargo/git/checkouts/p2panda-fb44b351afe3c13e/103a133/p2panda-rs/src/test_utils/db/test_db.rs:250:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

2.

```
---- db::stores::entry::tests::get_lipmaa_link_entries stdout ----
-------------- TEST START --------------
thread 'db::stores::entry::tests::get_lipmaa_link_entries' panicked at 'Send to store: InsertionError(OperationId(Hash("0020936c7abffc9a92f7e12d5a4e4b4b0bd1bbfcdfa3440aba43ba15224fd07bcd6b")))', /home/adz/.cargo/git/checkouts/p2panda-fb44b351afe3c13e/103a133/p2panda-rs/src/test_utils/db/test_db.rs:250:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
3.

```
---- graphql::client::dynamic_query::test::single_query stdout ----
-------------- TEST START --------------
thread 'graphql::client::dynamic_query::test::single_query' panicked at 'assertion failed: `(left == right)`
  left: `Null`,
 right: `Object({Name("byViewId"): Object({Name("fields"): Object({Name("bool"): Boolean(true)})}), Name("byDocumentId"): Object({Name("fields"): Object({Name("bool"): Boolean(true)})})})`: [
    ServerError {
        message: "Unknown field \"schema_name_0020e97f9999c4d8ab054d3dd4d0c1d8e4c7063cc06706594d69828dfae5fd22d6e9\" on type \"QueryRoot\".",
        locations: [
            Pos(2:17),
        ],
        path: [],
        extensions: None,
    },
    ServerError {
        message: "Unknown field \"schema_name_0020e97f9999c4d8ab054d3dd4d0c1d8e4c7063cc06706594d69828dfae5fd22d6e9\" on type \"QueryRoot\".",
        locations: [
            Pos(5:17),
        ],
        path: [],
        extensions: None,
    },
]', aquadoggo/src/graphql/client/dynamic_query.rs:440:13
```

sandreae:
On the off chance it's related, I'm gunna pull the e2e test out of this PR into it's own. :+1:

adzialocha:
> On the off chance it's related, I'm gunna pull the e2e test out of this PR into it's own. +1

Hmn .. none of these errors are related to the e2e test

sandreae:
> > On the off chance it's related, I'm gunna pull the e2e test out of this PR into it's own. +1
> 
> Hmn .. none of these errors are related to the e2e test

I know, but it seems like a new variable (running `Node` directly) which is worth removing :+1:

sandreae:
Booo, I just got similar errors:

```

failures:

---- graphql::client::dynamic_query::test::single_query stdout ----
-------------- TEST START --------------
thread 'graphql::client::dynamic_query::test::single_query' panicked at 'assertion failed: `(left == right)`
  left: `Null`,
 right: `Object({Name("byViewId"): Object({Name("fields"): Object({Name("bool"): Boolean(true)})}), Name("byDocumentId"): Object({Name("fields"): Object({Name("bool"): Boolean(true)})})})`: [
    ServerError {
        message: "Unknown field \"schema_name_0020ed39a1975cc0dfb369cb54bcd662ebf29c4563af22cfc8feb578eb5ba58f9b13\" on type \"QueryRoot\".",
        locations: [
            Pos(2:17),
        ],
        path: [],
        extensions: None,
    },
    ServerError {
        message: "Unknown field \"schema_name_0020ed39a1975cc0dfb369cb54bcd662ebf29c4563af22cfc8feb578eb5ba58f9b13\" on type \"QueryRoot\".",
        locations: [
            Pos(5:17),
        ],
        path: [],
        extensions: None,
    },
]', aquadoggo/src/graphql/client/dynamic_query.rs:440:13

---- graphql::client::dynamic_types::tests::application_schema_container_type stdout ----
-------------- TEST START --------------
thread 'graphql::client::dynamic_types::tests::application_schema_container_type' panicked at 'assertion failed: `(left == right)`
  left: `Object({Name("schema"): Null})`,
 right: `Object({Name("schema"): Object({Name("kind"): String("OBJECT"), Name("name"): String("schema_name_0020901c9ab87fa9da3e44ce56a507bcd0ff5075354bb3acdf663bbe540459a11610"), Name("description"): String("test schema description"), Name("fields"): List([Object({Name("name"): String("meta"), Name("type"): Object({Name("name"): String("DocumentMeta")})}), Object({Name("name"): String("fields"), Name("type"): Object({Name("name"): String("schema_name_0020901c9ab87fa9da3e44ce56a507bcd0ff5075354bb3acdf663bbe540459a11610Fields")})})])})})`: 
[]
', aquadoggo/src/graphql/client/dynamic_types/tests.rs:134:9

---- graphql::replication::query::tests::entries_newer_than_seq_num_cursor::case_3_some_edges_no_next_page stdout ----
-------------- TEST START --------------
thread 'graphql::replication::query::tests::entries_newer_than_seq_num_cursor::case_3_some_edges_no_next_page' panicked at 'Send to store: Custom("Entry's claimed log id of 1 does not match existing log id of 0 for given author and document")', /home/sandreae/.cargo/git/checkouts/p2panda-fb44b351afe3c13e/103a133/p2panda-rs/src/test_utils/db/test_db.rs:250:26


failures:
    graphql::client::dynamic_query::test::single_query
    graphql::client::dynamic_types::tests::application_schema_container_type
    graphql::replication::query::tests::entries_newer_than_seq_num_cursor::case_3_some_edges_no_next_page
```
Which vanished on a second run :sob:

sandreae:
(That was after removing the e2e test)

sandreae:
When it happens, it seems likely `graphql::client::dynamic_query::test::single_query` will be in there.

sandreae:
Hmm, every test I've seen fail has been using the `gql` interface.

sandreae:
Maybe increasing the `broadcast::channel` capacity higher that `16` is making a difference....

sandreae:
> Maybe increasing the `broadcast::channel` capacity higher that `16` is making a difference....

Fingers crossed, I think that was it. `16` seemed to be too low. 

Also in the `e2e` tests, when running the tests with `RUST_LOG=trace` we need to wait longer after publishing an entry for the node to process it.

aquadoggo/src/db/provider.rs
sandreae: Will do this in follow-up PR when I make some other storage provider api changes.

aquadoggo/src/db/utils.rs
sandreae: Raised an issue: https://github.com/p2panda/aquadoggo/issues/244

aquadoggo/src/domain.rs
adzialocha: Yes, we can remove it. The only thing we loose is a nicer error message.

aquadoggo/src/replication/service.rs
```diff
@@ -136,14 +140,13 @@ async fn verify_entries(entries: &[StorageEntry], context: &Context) -> Result<(
...
-                entry
-                    .operation_encoded()
-                    .map(|operation| operation.to_bytes()),
+                entry.into_bytes(),
+                entry.payload().map(|payload| payload.into_bytes()),
             )
         })
         .collect();
 
+    // @TODO: Do we want to use our own verification here?
```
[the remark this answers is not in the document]
adzialocha: Yes, lets make a ticket for that!

aquadoggo/src/tests.rs
```diff
@@ -0,0 +1,402 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+//! E2E test and mini-tutorial for aquadoggo.
```
adzialocha, line 3: This is very beautiful! :rose: 

I've renamed the file to `tests.rs` as the place of the file in the root already indicates that it is "end-to-end".

adzialocha (approved):
Yesssssss, this is the final step and it looks great! おつかれさまでした!

I think all other things can be easily tackled in sub-sequent steps, for now its just important that we have everything in place :sparkles: 

Small bummer: I continue having some failing cases when running tests .. much more rare now though.

```
---- graphql::client::dynamic_types::tests::application_schema_fields_type stdout ----
-------------- TEST START --------------
thread 'graphql::client::dynamic_types::tests::application_schema_fields_type' panicked at 'assertion failed: `(left == right)`
  left: `Object({Name("schemaFields"): Null})`,
 right: `Object({Name("schemaFields"): Object({Name("description"): String("Data fields available on documents of this schema."), Name("fields"): List([Object({Name("name"): String("bool_field"), Name("type"): Object({Name("kind"): String("SCALAR"), Name("name"): String("Boolean")})}), Object({Name("name"): String("list_field"), Name("type"): Object({Name("kind"): String("LIST"), Name("name"): Null})}), Object({Name("name"): String("relation_field"), Name("type"): Object({Name("kind"): String("OBJECT"), Name("name"): String("schema_definition_v1")})})])})})`:
[]
', aquadoggo/src/graphql/client/dynamic_types/tests.rs:216:9
```

sandreae:
👏👏👏👏🥳🥳🥳

Issue #247, opened 2022-08-19
Race conditions with failing tests


adzialocha:
I receive random failing tests when I run `cargo test` for the first times (very much looking like race conditions) :cry: - at one point they all pass ..

1. First run

```
---- db::stores::entry::tests::paginated_log_entries::case_2 stdout ----
-------------- TEST START --------------
thread 'db::stores::entry::tests::paginated_log_entries::case_2' panicked at 'Send to store: Custom("Entry's claimed seq num of 1 does not match expected seq num of 4 for given author and log")', /home/adz/.cargo/git/checkouts/p2panda-fb44b351afe3c13e/103a133/p2panda-rs/src/test_utils/db/test_db.rs:250:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

2. Second run

```
---- db::stores::entry::tests::get_lipmaa_link_entries stdout ----
-------------- TEST START --------------
thread 'db::stores::entry::tests::get_lipmaa_link_entries' panicked at 'Send to store: InsertionError(OperationId(Hash("0020936c7abffc9a92f7e12d5a4e4b4b0bd1bbfcdfa3440aba43ba15224fd07bcd6b")))', /home/adz/.cargo/git/checkouts/p2panda-fb44b351afe3c13e/103a133/p2panda-rs/src/test_utils/db/test_db.rs:250:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

3. Third run

```
---- graphql::client::dynamic_query::test::single_query stdout ----
-------------- TEST START --------------
thread 'graphql::client::dynamic_query::test::single_query' panicked at 'assertion failed: `(left == right)`
  left: `Null`,
 right: `Object({Name("byViewId"): Object({Name("fields"): Object({Name("bool"): Boolean(true)})}), Name("byDocumentId"): Object({Name("fields"): Object({Name("bool"): Boolean(true)})})})`: [
    ServerError {
        message: "Unknown field \"schema_name_0020e97f9999c4d8ab054d3dd4d0c1d8e4c7063cc06706594d69828dfae5fd22d6e9\" on type \"QueryRoot\".",
        locations: [
            Pos(2:17),
        ],
        path: [],
        extensions: None,
    },
    ServerError {
        message: "Unknown field \"schema_name_0020e97f9999c4d8ab054d3dd4d0c1d8e4c7063cc06706594d69828dfae5fd22d6e9\" on type \"QueryRoot\".",
        locations: [
[...]

PR #266, opened 2022-11-28
Debug failing tests

sandreae:
closes: #247 

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/materializer/service.rs
+++ b/aquadoggo/src/materializer/service.rs
@@ -213,9 +213,6 @@ mod tests {
             tx.send(crate::bus::ServiceMessage::NewOperation(first_operation_id))
                 .unwrap();
 
-            // Wait a little bit for work being done ..
-            tokio::time::sleep(Duration::from_millis(100)).await;
-
             // Make sure the service did not crash and is still running
             assert_eq!(handle.is_finished(), false);
 
@@ -281,10 +278,6 @@ mod tests {
             if rx_ready.await.is_err() {
                 panic!("Service dropped");
             }
-
-            // Wait for service to be done .. it should materialize the document since it was waiting
-            // as a "pending" task in the database
-            tokio::time::sleep(Duration::from_millis(200)).await;
 
             // Make sure the service did not crash and is still running
             assert_eq!(handle.is_finished(), false);
@@ -352,9 +345,6 @@ mod tests {
             ))
             .unwrap();
 
-            // Wait a little bit for work being done ..
-            tokio::time::sleep(Duration::from_millis(100)).await;
-
             // Then straight away publish an UPDATE on this document and send it over the bus too.
             let (entry_encoded, _) = send_to_store(
                 &db.store,
@@ -381,9 +371,6 @@ mod tests {
                 entry_encoded.hash().into(),
             ))
             .unwrap();
-
-            // Wait a little bit for work being done ..
-            tokio::time::sleep(Duration::from_millis(100)).await;
 
             // Make sure the service did not crash and is still running
             assert_eq!(handle.is_finished(), false);
@@ -455,9 +442,6 @@ mod tests {
                 p2panda_rs::entry::traits::AsEncodedEntry::hash(&entry_encoded).into(),
             ))
             .unwrap();
-
-            // Wait a little bit for work being done ..
-            tokio::time::sleep(Duration::from_millis(200)).await;
 
             // Make sure the service did not crash and is still running
             assert_eq!(handle.is_finished(), false);
--- a/aquadoggo/src/graphql/schema.rs
+++ b/aquadoggo/src/graphql/schema.rs
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
 //! Build and manage a GraphQL schema including dynamic parts of the schema.
+use core::time;
 use std::sync::Arc;
+use std::thread;
 
 use async_graphql::{EmptySubscription, MergedObject, Request, Response, Schema};
 use log::{debug, info};
--- a/aquadoggo/src/tests.rs
+++ b/aquadoggo/src/tests.rs
@@ -243,9 +243,6 @@ async fn publish(client: &Client, key_pair: &KeyPair, operation: &Operation) -> 
     // Which we post to the node.
     let _ = post(client, &query_str).await;
 
-    // Wait a little time for materialisation to happen.
-    tokio::time::sleep(Duration::from_millis(300)).await;
-
     encoded_entry.hash().into()
 }
 


aquadoggo/src/materializer/service.rs
```diff
@@ -213,8 +227,11 @@ mod tests {
             tx.send(crate::bus::ServiceMessage::NewOperation(first_operation_id))
                 .unwrap();
 
-            // Wait a little bit for work being done ..
-            tokio::time::sleep(Duration::from_millis(100)).await;
+            // Wait for the document to be materialised.
+            while !matches!(
+                rx_status.recv().await.unwrap(),
+                ServiceStatusMessage::Materialiser(TaskStatus::Completed(_))
+            ) {}
```
adzialocha, line 234: Wondering of this could be expressed a little differently, maybe with the `select` macro of tokio? Or we make it an own function?

sandreae: I can'd find a way to do it with `select`, with a `loop` we could do it like:

```
loop {
    if let ServiceStatusMessage::Materialiser(TaskStatus::Completed(_)) =
        rx_status.recv().await.unwrap()
    {
        break;
    }
}

```

adzialocha: Did ypu check the chapter "Pattern Matching" here? https://tokio.rs/tokio/tutorial/select

Otherwise we could refactor this too a simpler messaging format, similar to "trigger". The message is currently binary: pending / complete, it could be simply expressed by something which fires or not.

adzialocha: This verbose expression only exists because there is a `Pending` type, which is nowhere used, this is why we have to loop over it, otherwise recv would already be enough.

sandreae: We could introduce a new (more compact) message type instead of re-using `TaskStatus` and we could only send messages from the materializer service when a task completes. We don't really need do know about pending tasks here, that's more a debugging concern.

This would in turn simplify the wait/matching patterns required in simple cases.

sandreae: Maybe we don't need to know about dependency tasks at all either...

aquadoggo/Cargo.toml
```diff
@@ -39,6 +39,7 @@ p2panda-rs = { version = "^0.6.0", features = [
     "storage-provider",
 ] }
 serde = { version = "^1.0.144", features = ["derive"] }
+serial_test = "0.9.0"
```
sandreae, line 42: Need to move this into dev dependencies.

aquadoggo/Cargo.toml
```diff
@@ -78,5 +78,6 @@ reqwest = { version = "^0.11.11", default-features = false, features = [
 rstest = "^0.15.0"
 rstest_reuse = "^0.3.0"
 serde_json = "^1.0.85"
+serial_test = "0.9.0"
```
adzialocha, line 81: Let's pin this version by adding a `^` in front of it

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- State messaging channel for all services [#266](https://github.com/p2panda/aquadoggo/pull/266) 
+
+### Fixed
+
+- Fix polution of mutably shared schema store during testing [#266](https://github.com/p2panda/aquadoggo/pull/266)
```
adzialocha, line 16: "polution" typo

aquadoggo/src/bus.rs
```diff
@@ -13,3 +14,14 @@ pub enum ServiceMessage {
...
 }
+
+/// Messages sent on the service status channel.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum ServiceStatusMessage {
+    /// Message from the http service announcing the GraphQL Schema has been
+    /// built or re-built due to a new schema being materialized.
+    GraphQLSchemaBuilt,
+
+    /// Message from the materializer service containing a newly completed tasks.
```
adzialocha, line 25: "tasks" typo

aquadoggo/src/materializer/service.rs
```diff
@@ -115,9 +120,9 @@ pub async fn materializer_service(
         }
     });
 
-    debug!("Materialiser service is ready");
+    debug!("Materializer service is ready");
```
adzialocha, line 123: Hui, didn't we decide to write it with "s" 🤣

aquadoggo/src/materializer/service.rs
```diff
@@ -472,4 +486,87 @@ mod tests {
...
+                ServiceStatusMessage::MaterializerTaskComplete(task_2)
+            );
+
+            // // Wait for the document to be materialised.
+            // loop {
+            //     let msg = rx_status.recv().await.unwrap();
+            //     if let ServiceStatusMessage::Materializer(TaskStatus::Completed(task)) = msg {
+            //         break task;
+            //     }
+            // };
```
adzialocha, line 566: Why is this commented here?

adzialocha (commented):
Cool!

Small comments for now, will look at it a little closer soon as I'm looking at it from my phone only: Wondering if we could a) give our two "status channels" more distinctive meaning and b) make it more ergonomic to use our API.

Also I'd like to look again at the underlying problem, just to understand if it's really expected behavior and not a bug (so far I think it's not).

aquadoggo/src/materializer/service.rs
sandreae: Yeh, we did, but I didn't wanna change module names and code in unrelated places in this PR, so for consistency kept the status quo.

adzialocha:
@sandreae, after some experimentation I thought it might be a good idea to keep this PR open for now and rather only merge the parts which fix the race conditions. You can find it here: https://github.com/p2panda/aquadoggo/pull/269 - What do you think? Is this a good idea?

I think it will make us get to what we want faster: Fix the annoying race conditions. The newly introduced feature very helpful and definitely more elegant, but is maybe not "solved" yet as we're both not happy with its API design. It doesn't give us much functionality outside of tests but already affects the API for other non-tests parts of the stack, making them harder to grasp. I'd rather merge the important bit for now and keep this open until we find something pretty!

We can keep this here open and make a new ticket which deals with introducing that feature?

In any case, thanks so much for digging into all of that! What a ride!

adzialocha:
In this branch I sadly had still the following failing tests (as I've mentioned on the chat):

```
thread 'db::stores::entry::tests::paginated_log_entries::case_6' panicked at 'Send to store: Custom("Entry's claimed seq num of 1 does not match expected seq num of 23 for given public key and log")', /home/adz/.cargo/registry/src/github.com-1ecc6299db9ec823/p2panda-rs-0.6.0/src/test_utils/db/test_db.rs:246:26
```

```
thread 'graphql::client::dynamic_query::test::collection_query' panicked at 'assertion failed: `(left == right)`
  left: `Null`,
 right: `Object({Name("collection"): List([Object({Name("fields"): Object({Name("bool"): Boolean(true)})})])})`: [
    ServerError {
        message: "Unknown field \"all_schema_name_0020d0209be7d13c3ac6d464b337d07eb6dda7be72ef4943348cea2e95a111e3d05c\" on type \"QueryRoot\".",
        locations: [
            Pos(2:17),
        ],
        path: [],
        extensions: None,
    },
]', aquadoggo/src/graphql/client/dynamic_query.rs:604:13
```

sandreae:
> @sandreae, after some experimentation I thought it might be a good idea to keep this PR open for now and rather only merge the parts which fix the race conditions. You can find it here: #269 - What do you think? Is this a good idea?
> 
> I think it will make us get to what we want faster: Fix the annoying race conditions. The newly introduced feature very helpful and definitely more elegant, but is maybe not "solved" yet as we're both not happy with its API design. It doesn't give us much functionality outside of tests but already affects the API for other non-tests parts of the stack, making them harder to grasp. I'd rather merge the important bit for now and keep this open until we find something pretty!
> 
> Leaving your feature changes unmerged still keeps the second class of race conditions unfixed, but we can at least "control" them with bumping the waiting times. Also not beautiful, but it would come with no big changes for now.
> 
> We can keep this here open and make a new ticket which deals with introducing that feature? If you're happy with this I open #269 for review and create the ticket.
> 
> In any case, thanks so much for digging into all of that! What a ride!

Hey! Thanks for looking into this. I think what you're suggesting is a good move, it's a shame we might not use a lot of the code I did here, but maybe it was unavoidable, in any case it's good research which will feed into the next design phase!

sandreae:
> In this and the new branch I sadly had still the following failing test (as I've mentioned on the chat):
> 
> ```
> thread 'db::stores::entry::tests::paginated_log_entries::case_6' panicked at 'Send to store: Custom("Entry's claimed seq num of 1 does not match expected seq num of 23 for given public key and log")', /home/adz/.cargo/registry/src/github.com-1ecc6299db9ec823/p2panda-rs-0.6.0/src/test_utils/db/test_db.rs:246:26
> ```

Hmm... interesting.... let me see if I can reproduce it.

sandreae:
> > In this and the new branch I sadly had still the following failing test (as I've mentioned on the chat):
> > ```
> > thread 'db::stores::entry::tests::paginated_log_entries::case_6' panicked at 'Send to store: Custom("Entry's claimed seq num of 1 does not match expected seq num of 23 for given public key and log")', /home/adz/.cargo/registry/src/github.com-1ecc6299db9ec823/p2panda-rs-0.6.0/src/test_utils/db/test_db.rs:246:26
> > ```
> 
> Hmm... interesting.... let me see if I can reproduce it.

Tried reproducing this but so far never seen a fail, ran it 20x or so.... :disappointed:

sandreae:
Let's move this to the new PR though :+1:

adzialocha:
I've created two new issues to tackle the separate topics:

* https://github.com/p2panda/aquadoggo/issues/271
* https://github.com/p2panda/aquadoggo/issues/270

adzialocha:
I've closed this PR for now (keeping the `branch` intact if we still need it). One reason is that it is stale, the other is that I think we can tackle this issue from another perspective: Rather than changing our code for better tests we can introduce new test utilities.

My suggestion would be to have a "timeout helper" which checks for a condition every x milliseconds and times out if that condition wasn't reached after y seconds. Like that we can test async conditions to arrive eventually. I've seen that pattern in other projects as well and think that this is the way to go for at least test-related problems.

PR #269, opened 2023-01-24
Fix test-runner race conditions

adzialocha:
Some of our tests use the underlying static schema provider which is a static mutable data store (due to a workaround for `async-graphql`). This mutable store is accessible across all test runner threads in parallel mode as they all run in the same process. To prevent overwriting data across threads (race condition) we have to run these test in serial.

Read more: https://users.rust-lang.org/t/static-mutables-in-tests/49321

## Background

This PR is blantantly stolen from https://github.com/p2panda/aquadoggo/pull/266 which is the in-depth research of @sandreae to find out where the race-conditions came from.

The following changes have been made in comparison to PR #266:

1. Added comments around the places where we add `[#serial]`. It's fairly special-case and I can see how that same problem will come up in the future when we (and others) add new tests and forget to make them execute in serial. These comments will not necessarily prevent that but it's better than nothing maybe.
2. Removed all code which was somewhat _optional_ around solving these kinds of race conditions. #266 adds a feature to reliably inform about if an async task (like building a new schema) was finished. There is more race conditions to tackle, but they are more unlikely to happen (for now we can naively bump the waiting times).

I'm wondering if there's another way to tackle the issue in the future. We could have a mocked static schema store only for test environments which puts the data behind a randomized key which is set once per thread. Then all parallelly-running tests would have the schema store behind the same memory address, but access different areas _inside_ of it.

Closes: #247

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix pollution of mutably shared schema store during testing
+
 ## [0.4.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix pollution of mutably shared schema store during testing
+- Fix race-condition of mutably shared static schema store during testing [#269](https://github.com/p2panda/aquadoggo/pull/269)
 
 ## [0.4.0]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -80,6 +80,7 @@ dependencies = [
  "rstest_reuse",
  "serde",
  "serde_json",
+ "serial_test",
  "sqlx",
  "thiserror",
  "tokio",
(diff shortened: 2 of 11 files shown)


adzialocha:
Still observing the following failing tests:

1.

```
thread 'db::stores::entry::tests::entry_by_seq_number' panicked at 'Send to store: Custom("Entry's claimed log id of 1 does not match existing log id of 0 for given public key and document")', /home/adz/.cargo/registry/src/github.com-1ecc6299db9ec823/p2panda-rs-0.6.0/src/test_utils/db/test_db.rs:246:26
```

2. 

```
thread 'db::stores::entry::tests::paginated_log_entries::case_1' panicked at 'Send to store: Custom("Entry's claimed log id of 1 does not match existing log id of 0 for given public key and document")', /home/adz/.cargo/registry/src/github.com-1ecc6299db9ec823/p2panda-rs-0.6.0/src/test_utils/db/test_db.rs:246:26
```

sandreae (approved):
Looks good to me!

You mentioned still seeing one test fail sometimes [here](https://github.com/p2panda/aquadoggo/pull/266#issuecomment-1402437372) but I couldn't reproduce this myself.

adzialocha:
> Looks good to me!
> 
> You mentioned still seeing one test fail sometimes [here](https://github.com/p2panda/aquadoggo/pull/266#issuecomment-1402437372) but I couldn't reproduce this myself.

Thanks! I'll make an issue about it and then we keep looking out for it!

Issue #278, opened 2023-02-20
Build and configure `libp2p` swarm for inter-node communication


mycognosist:
First steps toward `libp2p` service integration for networking.

- Define swarm configuration parameters (listening addresses, max connections etc.)
  - Expose optional private network configuration (with key) 
- Build the swarm using QUIC transports
- Log swarm events
- Define the `NetworkBehaviour` of the node
  - `keep_alive` and `ping`
  - `mDNS`
  - Dummy replication (self-implemented behaviour for learning purposes)

I'm building on the `libp2p` branch started by @sandreae : https://github.com/p2panda/aquadoggo/tree/libp2p

mycognosist:
Base implementation PR: https://github.com/p2panda/aquadoggo/pull/282

PR #282, opened 2023-02-28
Add libp2p service and configuration

mycognosist:
Implements a minimal networking layer using the `libp2p` library, including transport establishment, networking configuration and network behaviour composition.

Relates to https://github.com/p2panda/aquadoggo/issues/278

## To Do

- [x] Keypair configuration
  - [x] Generate a new keypair when one doesn't exist
  - [x] Persist the keypair to file using PEM format
  - [x] Load the keypair from file 
- [x] Define swarm configuration parameters
  - [x] Set connection-related defaults 
  - [x] Decide on default QUIC port
- [x] Define the NetworkBehaviour of the node
  - [x] ping
  - [x] mDNS
- [x] Swarm
  - [x] Configure and create QUIC transport  
  - [x] Listen for incoming connections
  - [x] Dial peer addresses for replication (if provided)
  - [x] Log events
- [x] Install Protocol Buffers compiler for CI workflows
- [x] Add Protocol Buffers compiler installation requirement note to `README`

## Notes

`aquadoggo` currently uses `envy` to parse environment variables and overlay them onto the default configuration parameters. However, it does not currently handle nested `struct`s. This means that it is not possible to tune libp2p configuration parameters given the current code structure (the same applies to `ReplicationConfig`). I will need to use the `flatten` attribute from `serde` ([docs](https://serde.rs/attr-flatten.html)) to achieve this.

## Out of Scope

I'm noting this down here for future iterations on the networking layer functionality.

- Dummy replication (self-implemented behaviour for learning purposes)
  - Let's rather do this after the base networking functionality has been merged
- Address book
  - Read from file
  - Add discovered peers (via mDNS, for example)
  - Write to file
- `TOML` configuration file for the application
  - I'll tackle this in another PR

## 📋 Checklist


mycognosist:
Looks like protocol buffers will need to be installed in the CI.

sandreae:
This is looking great, with the networking features added by this module it `aquadoggo` really starts to feel like a sociable little creature :-) can't wait to start playing around with it.

sandreae:
I think something funky happened with the rebase, i can see we still have history here we don't want.... Anyway, i'll sort that later.

mycognosist:
Alrighty @adzialocha and @sandreae , I reckon this is ready for review. There's still some polishing that can be done but I'd rather not burn hours prematurely. I've left the tests tick-box unchecked for now because I'd like your input.

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+use log::info;
+use serde::{Deserialize, Serialize};
+
+use crate::libp2p::identity::Identity;
+
+/// Keypair file name.
+const KEYPAIR_FILE_NAME: &str = "libp2p.pem";
+
+/// QUIC default transport port.
+const QUIC_PORT: u16 = 48648;
```
mycognosist, line 18: This number is chosen from the list of unassigned ports (as defined by IANA).

aquadoggo/src/libp2p/identity.rs
```diff
@@ -0,0 +1,111 @@
...
+        PeerId::from(self.public())
+    }
+
+    /// Return the keypair.
+    fn keypair(&self) -> Keypair {
+        self.clone()
+    }
+
+    /// Encode the keypair using the PEM (Privacy-Enhanced Mail) format.
+    fn encode_pem(&self) -> String {
```
mycognosist, line 45: I chose to use an existing key-storage format rather than chucking the keys into a `TOML` file (or similar).

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+
+/// Libp2p config for the node.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct Libp2pConfiguration {
+    /// Dial concurrency factor.
+    ///
+    /// Number of addresses concurrently dialed for a single outbound
+    /// connection attempt.
+    pub dial_concurrency_factor: u8,
```
adzialocha, line 27: First I thought this parameter is about "Maximum outgoing connections." but then I realised it is something else, I'm not sure though what. "An attempt to create ONE outgoing connection by using MULTIPLE addresses?". What is dialing in this context? Is there a different way to describe what this config changes?

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub max_connections_in: u32,
+
+    /// Maximum outgoing connections.
+    pub max_connections_out: u32,
+
+    /// Maximum pending incoming connections.
+    pub max_connections_pending_in: u32,
+
+    /// Maximum pending outgoing connections.
+    pub max_connections_pending_out: u32,
```
adzialocha, line 42: What is a "pending" connection? Is this some sort of waiting queue?

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    /// Connection event buffer size.
+    ///
+    /// Defines the buffer size for events sent from the ConnectionHandler to the
+    /// NetworkBehaviour. Each connection has its own buffer. If the buffer is
+    /// exceeded, the ConnectionHandler will sleep.
+    pub per_connection_event_buffer_size: usize,
+
+    /// Ping behaviour enabled.
+    pub ping: bool,
```
adzialocha, line 65: Maybe we can add some notes here what this behaviour enables.

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub max_connections_pending_in: u32,
+
+    /// Maximum pending outgoing connections.
+    pub max_connections_pending_out: u32,
+
+    /// Maximum connections per peer (includes outgoing and incoming).
+    pub max_connections_per_peer: u32,
+
+    /// mDNS discovery enabled.
+    pub mdns: bool,
```
adzialocha, line 48: Maybe we can add some notes here what this behaviour enables.

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// If the buffer is exceeded, the Swarm will have to wait. An individual buffer with this
+    /// number of events exists for each individual connection.
+    pub notify_handler_buffer_size: usize,
+
+    /// Connection event buffer size.
+    ///
+    /// Defines the buffer size for events sent from the ConnectionHandler to the
+    /// NetworkBehaviour. Each connection has its own buffer. If the buffer is
+    /// exceeded, the ConnectionHandler will sleep.
+    pub per_connection_event_buffer_size: usize,
```
adzialocha, line 62: These notes are great but maybe slightly too libp2p specific? Is there a way to talk about what they change without going too much into specific struct / traits? Thinking about an `aquadoggo` crate consumer who just wants to configure the thing without needing to know how it works internally.

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        Self {
+            dial_concurrency_factor: 8,
+            listening_multiaddr,
+            max_connections_in: 16,
+            max_connections_out: 16,
+            max_connections_pending_in: 8,
+            max_connections_pending_out: 8,
+            max_connections_per_peer: 8,
+            mdns: false,
```
adzialocha, line 83: I like that all behaviour is turned off by default, for the `aquadoggo_cli` I can imagine we want it to be turned on by overwriting that default.

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn connection_limits(&self) -> ConnectionLimits {
+        ConnectionLimits::default()
+            .with_max_pending_outgoing(Some(self.max_connections_pending_out))
+            .with_max_pending_incoming(Some(self.max_connections_pending_in))
+            .with_max_established_outgoing(Some(self.max_connections_out))
+            .with_max_established_incoming(Some(self.max_connections_in))
+            .with_max_established_per_peer(Some(self.max_connections_per_peer))
+    }
+
+    /// Load the keypair from the file at the specified path.
```
adzialocha, line 102: Extra newline here, then it will render nicely in RustDoc previews :sparkles:

adzialocha: Super nitpick: We spell "key pair" or `key_pair` separated everywhere else. Something I notice because we talked about it for ages in the past haha

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// If the file does not exist, a random keypair is generated and saved.
+    /// If no path is specified, a random keypair is generated.
+    pub fn load_or_generate_keypair(path: Option<PathBuf>) -> Result<Keypair> {
```
adzialocha, line 105: Wow, nodes with key pairs :exploding_head:

adzialocha: .. soon I want a private GraphQL API where I can send operation data and the node encodes and signs the entries internally :laughing:

aquadoggo/src/libp2p/config.rs
```diff
@@ -0,0 +1,163 @@
...
+        };
+
+        Ok(keypair)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::Libp2pConfiguration;
+    use tempfile::TempDir;
```
adzialocha, line 137: Mini style nitpick: We organise imports in that order and with a newline in-between: 1. external 2. crate 3. super (while we're using super mostly only in tests).

aquadoggo/src/libp2p/identity.rs
adzialocha: Is there a way to store only the private key as an hex-encoded string? It would allow for easier access, especially if you might want to copy it over in another context (thats not really what we want users to do, but even just for debugging its helpful).

I'd prefer that and then convert it all towards libp2p in the way they want, but towards us a simple text file could be nice (similar to how `send-to-node` does it).

Optionally we can also use "our" identity API to generate the key pair: https://docs.rs/p2panda-rs/latest/p2panda_rs/identity/index.html - to have everything coming out of one hand?

aquadoggo/src/libp2p/identity.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /// Encode the keypair as PEM and write it to the given path.
+    fn save(&self, path: &Path) -> Result<()> {
+        let pem = self.encode_pem();
+
+        fs::create_dir_all(path.parent().unwrap())?;
+        let mut file = File::create(path)?;
+        file.write_all(pem.as_bytes())?;
+        file.sync_all()?;
```
adzialocha, line 79: Nitpick: A newline could be nice here to separate the writing from setting the permission a bit

aquadoggo/src/libp2p/service.rs
```diff
@@ -0,0 +1,195 @@
...
+use crate::bus::ServiceSender;
+use crate::context::Context;
+use crate::libp2p::Libp2pConfiguration;
+use crate::manager::{ServiceReadySender, Shutdown};
+
+/// Network behaviour for the aquadoggo node.
+#[derive(NetworkBehaviour)]
+struct Behaviour {
+    /// Automatically discover peers on the local network.
+    mdns: Toggle<mdns::tokio::Behaviour>,
```
adzialocha, line 23: Style nitpick: We usually make newlines between struct fields

aquadoggo/src/libp2p/service.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    if tx_ready.send(()).is_err() {
+        warn!("No subscriber informed about libp2p service being ready");
+    };
+
+    // Wait until we received the application shutdown signal or handle closed
+    tokio::select! {
+        _ = handle => (),
+        _ = shutdown => {
+        },
```
adzialocha, line 191: This could also be an empty tuple

aquadoggo_cli/src/main.rs
```diff
@@ -48,13 +52,22 @@ struct Opt {
...
     #[structopt(short = "A", parse(try_from_str = parse_key_val), number_of_values = 1)]
     public_keys_to_replicate: Vec<(String, Vec<u64>)>,
+
+    /// Enable mDNS for peer discovery over LAN.
+    #[structopt(short, long)]
+    mdns: bool,
+
+    /// Enable ping for connected peers (send and receive ping packets).
+    #[structopt(long)]
+    ping: bool,
```
adzialocha, line 62: As mentioned above I think these could be enabled by default in the client application?

adzialocha: Imagining people playing w. it (without reading the docs ..) and wondering why peers can't find each other "automagically" :laughing:

aquadoggo/src/libp2p/service.rs
```diff
@@ -0,0 +1,195 @@
...
+    // Subscribe to communication bus
+    let mut _rx = tx.subscribe();
+
+    // Read the libp2p configuration parameters from the application context
+    let libp2p_config = context.config.libp2p.clone();
+
+    // Load the libp2p keypair and peer ID
+    let keypair = Libp2pConfiguration::load_or_generate_keypair(context.config.base_path.clone())?;
+    let peer_id = PeerId::from(keypair.public());
+    info!("libp2p peer ID: {peer_id:?}");
```
adzialocha, line 73: I noticed that this prints as `libp2p peer ID: PeerId("12D3KooWQFPNhLX9h9xjwTEv1xvzkxzTh7PrTw5F82am3r5Xp5gq")` in the logs w. `Debug` (aka w. `PeerId()` type around value) representation. Is it possible to use the `Display` representation instead?

aquadoggo/src/libp2p/service.rs
```diff
@@ -0,0 +1,195 @@
...
+                    reason,
+                } => info!("ListenerClosed: {listener_id:?} {addresses:?} {reason:?}"),
+                SwarmEvent::ListenerError { listener_id, error } => {
+                    info!("ListenerError: {listener_id:?} {error:?}")
+                }
+                SwarmEvent::NewListenAddr {
+                    address,
+                    listener_id,
+                } => {
+                    info!("Listening on {address:?} {listener_id:?}");
```
adzialocha, line 172: Similar here: `Listening on "/ip4/127.0.0.1/udp/48648/quic-v1" ListenerId(2898880830502420286)` (`Debug` vs. `Display`)

aquadoggo/src/node.rs
```diff
@@ -84,6 +85,11 @@ impl Node {
             panic!("Failed starting HTTP service");
         }
 
+        // Start libp2p service
+        if manager.add("libp2p", libp2p_service).await.is_err() {
+            panic!("Failed starting libp2p service");
```
adzialocha, line 90: Similarily here we might want to give this service a different name here. People will look at it from the outside and wonder what this "libp2p" is doing

adzialocha (changes requested):
Woho! We have a running QUIC server in `aquadoggo` :sparkles: and key pairs! And mDNS discovery! :birthday: 

The whole code looks really good, I don't really have much to note here.

**Naming**

The main two points which came up for me were:

1. General naming of "libp2p". I'd suggest to not use the word libp2p as an identifier for what this code does, also not towards the outside (in logs or API). libp2p is too powerful to just be "one thing" and on top we can't assume that people will even know what it is. Think the API would explain itself better if it would be named around "connections" or similar. Later we should separate our code in even more distinct parts like: "connections", "gossip", "relay", "mdns", etc.
2. Mini style nitpicks

**Running**

I've tried to make two `aquadoggo` instances run on my machine so they can find each other via mDNS but failed. What is the recommended approach?

In the first terminal I started:

```
RUST_LOG=aquadoggo=debug cargo run -- --mdns
```

.. in the second:

```
RUST_LOG=aquadoggo=debug cargo run -- --mdns --http-port 2024
```

.. and it fails with:

```
[2023-03-06T10:30:17Z INFO  aquadoggo::manager] Start libp2p service
[2023-03-06T10:30:17Z INFO  aquadoggo::libp2p::service] libp2p peer ID: PeerId("12D3KooWQFPNhLX9h9xjwTEv1xvzkxzTh7PrTw5F82am3r5Xp5gq")
[2023-03-06T10:30:17Z ERROR aquadoggo::manager] Error in libp2p service:
thread 'main' panicked at 'Failed starting libp2p service', /home/adz/dev/p2panda/aquadoggo/aquadoggo/src/node.rs:90:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Defaults**

The dream behaviour would be to just start `aquadoggo` and it would magically find a free port and start talking mDNS without any further enabling of any flags or config. We can handle that in a separate PR though as this seems optional for now.

aquadoggo/src/libp2p/config.rs
adzialocha: Maybe the key pair file should be called `secret` or `private-key`?

mycognosist: Cool, I'll change to `2022` and `private-key`.

aquadoggo/src/config.rs
```diff
@@ -7,6 +7,7 @@ use anyhow::Result;
 use directories::ProjectDirs;
 use serde::Deserialize;
 
+use crate::libp2p::Libp2pConfiguration;
```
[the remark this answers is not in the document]
mycognosist: Yes, good idea to change the naming. My suggestion would be `network`. This is a pattern followed by quite a few projects I've looked at which use`libp2p` under the hood. It's also less specific than `connection` (in a good way). Then we'd have `NetworkConfiguration`.

aquadoggo/src/libp2p/config.rs
mycognosist: You are attempting to connect with a peer for whom you hold multiple addresses. This value defines the number of addresses you will dial concurrently in an attempt to make a connection with the peer.

aquadoggo/src/libp2p/config.rs
mycognosist: Yes, it's a waiting queue of sorts. A pending connection is one which has been initiated but has not yet received a response.

For example, once we  receive an inbound connection request it is added to the `pending inbound` queue. Once we respond to the request it moves out of the `pending` queue.

mycognosist:
@adzialocha 

Thanks for the thorough review! I'm working on the requested changes.

> I've tried to make two `aquadoggo` instances run on my machine so they can find each other via mDNS but failed. What is the recommended approach?

It fails because of a port conflict. mDNS uses port `5353`. The recommended approach is to run one instance per machine (or to somehow isolate the instances on a single machine).

mycognosist:
**Requested Changes**

_This is just an easier way for me to track progress on changes._

- [ ] `libp2p` -> `network`
- [ ] `Libp2pConfiguration` -> `NetworkConfiguration`
- [ ] Use `p2panda_rs/identity` to generate `libp2p` keypair
- [ ] `PEM` keyfile -> hex-encoded string
- [ ] `libp2p.pem` -> `private-key`
- [ ] QUIC port -> `2022`
- [ ] Enable `ping` and `mdns` by default in `aquadoggo_cli`
- [ ] Add doc comment notes (behaviour descriptions) for `ping` and `mdns`
- [ ] Make doc comments more generic for `notify_handler_buffer_size` and `per_connection_event_buffer_size`
- [ ] Address all style changes (newline, empty tuple, test dependency import order etc.)

aquadoggo/src/libp2p/service.rs
mycognosist: Unfortunately `ListenerId` does not currently implement `Display`.

adzialocha: Maybe we take it out then?

mycognosist: I have removed the logging of `ListenerId` for that event and have knocked the logging of most swarm events down to `debug`. That sounds make things less noisy while still providing valuable information for debugging.

adzialocha:
> @adzialocha 
> 
> It fails because of a port conflict. mDNS uses port `5353`. The recommended approach is to run one instance per machine (or to somehow isolate the instances on a single machine).

I see, thank you! Is this specified by libp2p? For multicast sockets it's usually an exception to re-bind to the same port.

mycognosist:
@adzialocha 

> I see, thank you! Is this specified by libp2p? For multicast sockets it's usually an exception to re-bind to the same port via `SO_REUSEADDR`.

Aha! Excellent question! This helped me to fix some of my incorrect assumptions. Thank you!

Going back to your failed attempt to run two nodes on your local machine with mDNS: I was wrong. The cause of the panic was actually the QUIC port!

I dug into the libp2p code to look for the port reuse option and ended up finding it [here](https://github.com/libp2p/rust-libp2p/blob/be3ec6c62b05b37f85fd60ce1a916f33235522af/protocols/mdns/src/behaviour/iface.rs#L86) (their mDNS docs are very sparse and do not mention this, or the port `5353`). Then I tried to run two `aquadoggo` nodes on my local machine, one with mDNS enabled and one with it disabled. That still resulted in a panic. That's when I thought it might be the QUIC port conflict. I added a `--quic-port` CLI option to test my assumption and it was correct; I now have mDNS working when one node is using `2022` (QUIC default) and the other is on `2023`.

Let me clean up the code and commit it.

P.s. I'm really sorry that this PR has become so big and complicated -_-

adzialocha (approved):
Looks great! ✨ Thank you for all the extra changes. I think it's good to go 👍🏻

sandreae (changes requested):
This is a great PR and represents a huge amount of very thorough research, thanks so much for guiding and grounding this complex area. We have a network!!!!

Just some small comments in this review, i agreed with all the points @adzialocha already made. One extra thing is that I wonder if we need some more info in the README specifically about the network layer. This can happen in a follow-up PR and once we have the other pieces in place.

Also, maybe i'm doing something wrong, but should i be getting any chatter when running two nodes locally? Like mDNS or ping? I don't see anything in the logs when i run:

`RUST_LOG=aquadoggo=debug cargo run`

and then:

`RUST_LOG=aquadoggo=debug cargo run -- --http-port 2021 --quic-port 2023`  

I'd just like to check this point and the small comments i made then I'm very happy to approve and merge.

adzialocha:
> Just some small comments in this review, i agreed with all the points @adzialocha already made. One extra thing is that I wonder if we need some more info in the README specifically about the network layer. This can happen in a follow-up PR and once we have the other pieces in place.

Agreed that it will be good to write something in the `README.md` but maybe that's still too early? I would write something if we have a fully fledged out networking layer, but so far that's "only the beginning" :laughing:

aquadoggo/Cargo.toml
```diff
@@ -31,6 +31,14 @@ futures = "^0.3.23"
 gql_client = "^1.0.6"
 hex = "^0.4.3"
 http = "^0.2.8"
+libp2p = { version = "^0.51.0", features = [
+    "tokio",
+    "noise",
```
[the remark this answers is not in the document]
mycognosist: Good catch! We do not in fact need `noise`. I'll remove it.

mycognosist:
@sandreae 

Thanks so much for the review! 🖤 Loads of awesome teamwork and sharing has made this possible.

I have removed the `noise` feature (well spotted!) and have added a note about `mDNS` to the network service doc comment.

> Also, maybe i'm doing something wrong, but should i be getting any chatter when running two nodes locally? Like mDNS or ping?

You need one small addition to your second invocation to specify a non-default base path. That results in a separate peer ID being created for the second node. The way you tried it before meant that both nodes were running with the same ID and therefore `mDNS` does not discover a _new_ node.

`RUST_LOG=aquadoggo=debug cargo run`

`RUST_LOG=aquadoggo=debug cargo run -- --http-port 2021 --quic-port 2023 --data-dir=$HOME/test`

You should see something like this in the log output:

`[2023-03-10T11:31:38Z DEBUG aquadoggo::network::service] mDNS discovered a new peer: 12D3KooWDC8Kmq9z9m4VSYZD68ioJYfxqR3DTKKfrjh7XMH9NXGZ`

Note that the `mDNS` protocol handler will need to be updated in a future PR to act on peer discovery events. At this point, the discovery event is simply logged. We'll want to either invoke `swarm.dial()` for the discovered peer directly in the handler or add the peer to the address book and handle the dialing elsewhere.

adzialocha:
> You need one small addition to your second invocation to specify a non-default base path. That results in a separate peer ID being created for the second node. The way you tried it before meant that both nodes were running with the same ID and therefore `mDNS` does not discover a _new_ node.

That's a new pattern: Nodes have key pairs now 🤯

sandreae:
Ah, thanks, yeh that did the trick! I see discovery happening now :-)

sandreae (approved):

PR #285, opened 2023-03-01
Do transactions correctly

sandreae:
closes: https://github.com/p2panda/aquadoggo/issues/283

## 📋 Checklist

- [x] Add tests that cover your changes
- [x] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [x] Link this PR to any issues it closes
- [x] New files contain a SPDX license header

Issue #281, opened 2023-02-28
Updates not getting materialized


adzialocha:
Sending a CREATE operation, followed by 1-2 UPDATE operations do _sometimes_ not get materialized.

**Steps to reproduce:**

1. Check out `document-orm` branch in `shirokuma`: https://github.com/p2panda/shirokuma/tree/document-orm
2. Run `aquadoggo` from `main` w. new database
3. Check out `send-to-node` from `debug-js` branch
4. Run `./debug.sh` in `send-to-node` (this creates a `chat` schema and adds a chat document with two updates, materialization seems to work normally here)
5. Go back to `shirokuma` project, run `node ./debug.js` here (tested in NodeJS v17)
6. Check GraphQl playground: The document should arrive at `message` value `3`, run step 5 a couple of times and compare, sometimes its `1`, sometimes its `2`, sometimes its `3`

adzialocha:
The entries and operations sent by `debug.js` look like that, should be all fine:

**1. CREATE**

Entry: `{ logId: '0', seqNum: '1', backlink: null, skiplink: null }`
Operation: `[1, 0, "chat_002009a6721c3015eb87a7b941e5d54183b7bd209c72dec05af234ede47dd4772f35", {"message": "1"}]`

**2. UPDATE**

Entry:

```
{
  logId: '0',
  seqNum: '2',
  backlink: '0020d089767fea4a13d8c5972fdabe8105b46d5da9da2bedef0c64b506d6a535b38e',
  skiplink: null
}
```

Operation: `[1, 1, "chat_002009a6721c3015eb87a7b941e5d54183b7bd209c72dec05af234ede47dd4772f35", ["0020d089767fea4a13d8c5972fdabe8105b46d5da9da2bedef0c64b506d6a535b38e"], {"message": "2"}]`

**3. UPDATE**

Entry:

```
{
  logId: '0',
  seqNum: '3',
  backlink: '00201089cea1f8c547f269dd490d9faae650a41dabfa74ce700cad0dad49ed7b58a8',
  skiplink: null
}
```

Operation: `[1, 1, "chat_002009a6721c3015eb87a7b941e5d54183b7bd209c72dec05af234ede47dd4772f35", ["00201089cea1f8c547f269dd490d9faae650a41dabfa74ce700cad0dad49ed7b58a8"], {"message": "3"}]`

adzialocha:
`aquadoggo` logs looking normal as well, just not showing the expected updates (sometimes):

```
[...]

PR #286, opened 2023-03-06
Introduce requeue flag

adzialocha:
Introduces a `requeue` boolean-flag in the input index of our `Factory` implementation (task worker queue) which allows tasks to get re-scheduled after completion.

This fixes a nasty bug which can be only reproduced on very fast machines: https://github.com/p2panda/aquadoggo/issues/281

This branches off Sam's work on https://github.com/p2panda/aquadoggo/pull/285 which surely made the bug easier to track down.

## Problem

Let's say we have operations arriving in the following order at the node: `C1 [D1]`, `U2 [D1]`, `C1 [D2]`, `U3 [D1]` and `U2 [D2]`. That's an CREATE, UPDATE and another UPDATE operation for a "Document 1" and another CREATE and UPDATE operation for a "Document 2", they arrive at the same time at the GraphQL API and get put in this order on the service bus where they get broadcasted.

The current factory implementation would now look at each of these operations arriving through the service bus and queue them up IF we don't handle that document yet. In the "dispatcher" this would look like that:

1. `C1 [D1]` arrives. Do we already work on `D1`? No, let's queue it!
2. `U2 [D1]` arrives. Do we already work on `D1`? Yes. Ignore.
3. `C1 [D2]` arrives. Do we already work on `D2`? No, let's queue it!
4. `U3 [D1]` arrives. Do we already work on `D1`? Yes. Ignore.
5. `U2 [D2]` arrives. Do we already work on `D2`? Yes. Ignore.

.. concurrently we're already running a whole worker pool eagerly waiting to look at the queue, taking the tasks off it. So depending on how the "dispatcher" races with the "workers" the outcome might always look a little bit different. For example, if we finished working on `C1 [D1]` _before_ the dispatcher looks at `U3 [D1]`, it might actually get queued as well.

That design of "looking at duplicates" had one purpose: Avoid working on the same document twice, aka "we're already working on it, why do it a second time". This is a nice optimization, but it also assumes that we have all possible operations in the database _already_ in the moment _before_ `C1 [D1]` or a later D1-related task kicks in.

Funnily, this worked quite well so far, we never noticed a problem. Most of the time the requests were slow enough for the factory to queue _every_ operation and ignore nothing:

1. `C1 [D1]` arrives. Do we already work on `D1`? No, let's queue it!
... (the worker happily crunches that operation in the background, seeing C1 in the database)
... some short fraction of time later ..
2. `U2 [D1]` arrives. Do we already work on `D1`? No, let's queue it!
... (the worker happily crunches that operation in the background, seeing U2 in the database)
... and so on ..

Since our requests get faster now (due to client-side caching in `shirokuma`) we observed the following problem:

1. `C1 [D1]` arrives. Do we already work on `D1`? No, let's queue it!
... (the worker happily crunches that operation in the background, seeing C1 in the database)
... some short fraction of time later ..
2. `U2 [D1]` arrives! Woah. FAST! Surprise!! Do we already work on `D1`? Yes. Ignore.
... and so on ..

The new operations `U2` etc. arrived in the database _after_ the worker looked at them but _before_ it finished working on `C1`! Thus, we lost `U2` ..

## Solution

After some analysis I realized that the system makes still sense in its basic assumption: We don't need to work on documents more than we have to! `aquadoggo` is an event-sourcing system where "events" kick-in work, independent of the amount of data arriving. Here is an example:

```
D: [U U U] [U U U U U U U] U U U U U U U U ...
          Work!           Work! 
```

Note how the brackets `[...]` "group" the operations in work units. They simply get loaded from the database as they already were there before the work started.

There is an ongoing stream of incoming UPDATE operations and every time we kick in the worker for that document it takes the "fresh" ones and materializes them. We don't need to have a worker for _each_ operation, we just have to make sure that we consider them all exactly _once_ - and this is where the previous system failed, as operations just got lost ..

In this PR a "re-queue" flag got introduced which restarts a task for a document D when operations have been observed which came in while the worker was already running on D. It works like this:

1. `C1 [D1]` arrives. Do we already work on `D1`? No, let's queue it!
2. `U2 [D1]` arrives. Do we already work on `D1`? Yes. Set requeue flag for `D1` to `true` 
3. `C1 [D2]` arrives. Do we already work on `D2`? No, let's queue it!
4. `U3 [D1]` arrives. Do we already work on `D1`? Yes, and the requeue flag is set. Ignore.
5. `U2 [D2]` arrives. Do we already work on `D2`? Yes. Set requeue flag for `D2` to `true` 

This scenario assumes that the worker for `D1` and the other worker for `D2` are still running while all other operations get dispatched, that's the "worst-case scenario" in terms of maximum pressure on the dispatcher which made the previous implementation fail.

What this does now is quite nice: As soon as the work on `C1` finished and the requeue flag was detected, it will just add a new task to the queue to continue working on `D1`, making sure to also account for all these operations which arrived a little bit later (`U2` and `U3`).

What we get from this is:

1. Some sort of "batching" of operations for the same document when they arrive _very_ fast
2. The queue does not get "blocked" by waiting operations, we can still continue looking at other operations for other documents and assigning them to different workers
3. Accounting for moments of very high pressure on the dispatcher without losing any information
4. Still making sure that we only working on a document _once_ at a time, to avoid concurrent work overwriting each other

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Correct use of `sqlx` transactions [#285](https://github.com/p2panda/aquadoggo/pull/285) `rs`
+- Correct use of `sqlx` transactions [#285](https://github.com/p2panda/aquadoggo/pull/285)
 - Fix race-condition of mutably shared static schema store during testing [#269](https://github.com/p2panda/aquadoggo/pull/269)
+- Introduce flag to requeue tasks in worker queue, fixes race-condition in materialization logic [#286](https://github.com/p2panda/aquadoggo/pull/286)
 
 ## [0.4.0]
 
--- a/aquadoggo/src/materializer/worker.rs
+++ b/aquadoggo/src/materializer/worker.rs
@@ -75,7 +75,7 @@
 //!
 //! Task 1 results in "25", Task 2 in "64", Task 4 in "9".
 //! ```
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::fmt::{Debug, Display};
 use std::future::Future;
 use std::hash::Hash;
@@ -139,6 +139,9 @@ pub enum TaskStatus<IN> {
 /// Workers are identified by simple string values.
 pub type WorkerName = String;
 
+/// Flag for the input index to indicate if we want to requeue the task after it completed.
+type RequeueFlag = bool;
+
 /// Every registered worker pool is managed by a `WorkerManager` which holds the task queue for
 /// this registered work and an index of all current inputs in the task queue.
 struct WorkerManager<IN>
@@ -149,7 +152,12 @@ where
     ///
     /// This allows us to avoid duplicate tasks by detecting if there is already a task in our
     /// queue with the same input hash.
-    input_index: Arc<Mutex<HashSet<IN>>>,
+    ///
+    /// An additional flag can be used to indicate that we want to requeue the same task again
+    /// after it completed. This is useful to account for more events which arrived _while_ the
+    /// task was processed. It is enough to only remember one of the potentially many events
+    /// arriving in this time, we're "batching" them for the next round.
+    input_index: Arc<Mutex<HashMap<IN, RequeueFlag>>>,
 
     /// FIFO queue of all tasks for this worker pool.
     queue: Arc<Queue<QueueItem<IN>>>,
@@ -162,7 +170,7 @@ where
     /// Returns a new worker manager.
     pub fn new() -> Self {
         Self {
-            input_index: Arc::new(Mutex::new(HashSet::new())),
+            input_index: Arc::new(Mutex::new(HashMap::new())),
             queue: Arc::new(Queue::new()),
         }
     }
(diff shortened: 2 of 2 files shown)


adzialocha:
`materialize_document_from_bus` seems to still be failing following the fixes @sandreae, I wonder whats hidden there :laughing:

sandreae:
This is amazing work. Sorry I only have the ability to respond in emoji reactions right now.

sandreae:
I just read through your great description of the new logic, it all makes sense , very elegant. Just to check I get it: the requeue flag is flipped to `true` if an operation arrives for a document `D1` while a task processing `D1` is in progress. This means a task for `D1` will be queued up again once the current one finishes. This means all operations will be accounted for, now concurrent work on the same document will occur, _and_ we don't inefficiently queue up a new task for every operation.

sandreae:
Gunna review now :+1:

sandreae (approved):
This looks all good to me, thanks for tracking down this tricky bug.

I never managed to reproduce this particular race condition on my (slow??) laptop, I'm assuming it fixed what you did manage to observe. 

Regarding the still failing test, I'd say we merge and look at that in a new PR.

sandreae:
Ah, I actually had a change request. Think it would be good to add debug logging for when an operation is requeued. I can add this, let me know if you have any logging related thoughts @adzialocha.

adzialocha:
> Just to check I get it: the requeue flag is flipped to true if an operation arrives for a document D1 while a task processing D1 is in progress. This means a task for D1 will be queued up again once the current one finishes. This means all operations will be accounted for, now concurrent work on the same document will occur, and we don't inefficiently queue up a new task for every operation.

I think you got the it, just maybe clarifying that "concurrent work on the same document will occur" is not true. There is no concurrent work on any same document allowed, otherwise we tap into undefined behaviour because of the database reads and writes on that document.

We also never queued up a new task for every operation. We queue up a task for every document one could say, the operation is the trigger for that. If a second operation for the same document comes in while we're already having a task, a "requeue" is flagged.

> I never managed to reproduce this particular race condition on my (slow??) laptop, I'm assuming it fixed what you did manage to observe.

You mean the `materialize_document_from_bus` test, right?

> Ah, I actually had a change request. Think it would be good to add debug logging for when an operation is requeued. I can add this, let me know if you have any logging related thoughts @adzialocha.

Sure, please go ahead!

adzialocha:
Thank you for rebasing!

sandreae:
> I think you got the it, just maybe clarifying that "concurrent work on the same document will occur" is not true. There is no concurrent work on any same document allowed, otherwise we tap into undefined behaviour because of the database reads and writes on that document.

Sorry, that was a typo on my part, should have read "no concurrent  work...."

sandreae:
> > > You mean the materialize_document_from_bus test, right?
> > 
> > 
> > No I mean, this queue bug, I couldn't reproduce it locally, it was only observable on your fast computer.
> 
> Ah, okay. I hope it's gone with this PR? laughing

But it's gone for you right? :sweat_smile:

PR #287, opened 2023-03-10
Use new version of `async-graphql` for dynamic schema generation

sandreae:
Use https://github.com/async-graphql/async-graphql and https://github.com/smmoosavi/dynamic-graphql to implement dynamic schema handling in our graphql api.

This took a little bit of time as I had to get to grips with the new `async-graphql` dynamic schema creation patterns, but I'm very happy with the result, it will be much more enjoyable (and quite quick) to expand our query api now with new features. 

## Design decisions
- we use `async-graphql` to build all queries dynamically. It doesn't seem possible to combine "statically" defined queries with dynamic ones, so even `nextArgs` is defined using the dynamic API
- all scalar and known types are defined using `dynamic-graphql` which offers macro wrappers for `async-graphql`, these are then registered on the root schema and can be referenced by name in dynamic queries where object types can be defined
- the resolver pattern i adopted (i believe it's the recommended way) is to have each field perform a single db query to retrieve it's own value. The query bubbles up from it's root to tips, only resolving actual values in it's final step. For this to be efficient, we may want to make two changes:
  1) implement a data loader which helps cache results for duplicate queries which occur in the same query tree (avoids N+1).
  2) write a new SQL query for the field resolver which only targets the requested field. 

## Tasks

- [x] remove old `graphql` module
- [x] remove replication end points (we might want these back later, but for now they're gone)
- [x] re-implement all functionality using above crates
  - [x] scalar types
  - [x] types 
    - [x] document meta 
    - [x] next args
  - [x] build document fields schema
  - [x] build document schema
  - [x] build document query one
  - [x] relations (single and lists)
  - [x] build document query all 
  - [x] build next args
  - [x] mutations: publish
- [x] bring back all tests
  - [x] make them pass.... 
- [x] ~add some more tests~ let's do this in next steps :face_in_clouds: we at least shouldn't have fewer tests than we did before!
- [x] ~get `dynamic-schema` published or remove it from this PR~ it already was

## Next steps
- implement a data loader for efficient query caching
- write field only SQL queries for field resolver (optional)

## 📋 Checklist


adzialocha:
We can probably remove all of the `serial` attributes in the tests (and maybe even that `test-serial` crate)?

aquadoggo/src/graphql/schema.rs
```diff
@@ -3,76 +3,112 @@
...
-/// GraphQL schema we can load from this store statically to build the schemas on the fly.
-async fn build_schema_with_workaround(shared: GraphQLSharedData) -> RootSchema {
-    // Store all application schemas from database into static in-memory storage
-    let all_schemas = shared.schema_provider.all().await;
-    save_static_schemas(&all_schemas);
-
-    // Build the actual GraphQL root schema, this will internally read the created JSON file and
-    // accordingly build the schema
-    build_root_schema(shared.store, shared.tx, shared.schema_provider)
+        .unwrap()
```
adzialocha, line 107: What could go wrong here when unwrapping?

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,474 @@
...
+use crate::graphql::utils::{downcast_id_params, fields_name, get_document_from_params};
+
+const QUERY_ALL_PREFIX: &str = "all_";
+const FIELDS_FIELD: &str = "fields";
+const META_FIELD: &str = "meta";
+const DOCUMENT_META_SCHEMA_ID: &str = "DocumentMeta";
+const DOCUMENT_ID_ARG: &str = "id";
+const DOCUMENT_VIEW_ID_ARG: &str = "viewId";
+const DOCUMENT_ID_SCALAR: &str = "DocumentId";
+const DOCUMENT_VIEW_ID_SCALAR: &str = "DocumentViewId";
```
adzialocha, line 21: Maybe add some documentation here what these constants are representing? I assume that the ones ending on "_SCALAR" need to match what is defined in the regarding `Scalar` trait implementations?

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,474 @@
...
+
+/// Build a graphql object type for a p2panda schema.
+///
+/// Contains resolvers for both `fields` and `meta`. The former simply passes up the query
+/// arguments to it's children query fields. The latter retrieves the document being queried and
+/// already constructs and returns the `DocumentMeta` object.
+pub fn build_document_schema(schema: &Schema) -> Object {
+    let document_fields_name = fields_name(schema.id());
+    Object::new(schema.id().to_string())
+        // The `fields` of a document, passes up the query arguments to it's children.
```
adzialocha, line 31: I stumbled upon this a little, maybe it should say "The `fields` field of a GraphQL document"? In general the GraphQL terminology mixes quite up with p2panda, but I guess there is not much we can do 😆

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,474 @@
...
+                    // Get the whole document.
+                    let document =
+                        get_document_from_params(store, &document_id, &document_view_id).await?;
+
+                    // Construct `DocumentMeta` and return it. We defined the document meta
+                    // type and already registered it in the schema. It's derived resolvers
+                    // will handle field selection.
+                    //
+                    // TODO: We could again optimize here by defining our own resolver logic
+                    // for each field.
```
adzialocha, line 62: I'm not sure if I understand this optimization idea. Do you suggest to have an individual SQL query for _each_ field? That wouldn't be optimal probably, or do you suggest to pass down the required list of fields to some sort of resolver which then builds _one_ custom SQL query based on the requested fields?

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,474 @@
...
+                    };
+
+                    Ok(field_value)
+                })
+            },
+        ))
+        .description(schema.description())
+}
+
+pub fn build_document_query(query: Object, schema: &Schema) -> Object {
```
adzialocha, line 81: A docstring could be nice here as well

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,474 @@
...
+                            );
+                        }
+                    };
+                    // Pass them up to the children query fields.
+                    Ok(Some(FieldValue::owned_any((document_id, document_view_id))))
+                })
+            },
+        )
+        // TODO: We'd be better off using an enum input type here i think, then we can specify it
+        // as required, which will provide a validation step for us.
```
adzialocha, line 133: How would that look like in the GraphQL query?

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,474 @@
...
+        .description(format!(
+            "Query a {} document by id or view id.",
+            schema.name()
+        )),
+    )
+}
+
+/// Add query for getting all documents of a certain schema type to the root query object.
+///
+/// Constructs an endpoint with the format `all_<SCHEMA_ID>`.
```
adzialocha, line 151: Is endpoint the right word in GraphQL land? Actually can't find really what else to mention. Would it be an Operation? (https://spec.graphql.org/October2021/#sec-Language.Operations)

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,474 @@
...
+
+    use crate::test_utils::{add_document, add_schema, graphql_test_client, test_runner, TestNode};
+
+    #[rstest]
+    // Note: This and more tests in this file use the underlying static schema provider which is a
+    // static mutable data store, accessible across all test runner threads in parallel mode. To
+    // prevent overwriting data across threads we have to run this test in serial.
+    //
+    // Read more: https://users.rust-lang.org/t/static-mutables-in-tests/49321
+    #[serial]
```
adzialocha: Did you try removing the `#[serial]` attributes? I think the tests should be able to run now concurrently as far as I remember the problem?

aquadoggo/src/graphql/schema_builders/next_args.rs
```diff
@@ -1,56 +1,76 @@
...
-            skiplink: skiplink.map(|hash| hash.into()),
+use crate::graphql::scalars::{DocumentViewIdScalar, PublicKeyScalar};
+use crate::graphql::types::NextArguments;
+
+const NEXT_ARGS_QUERY: &str = "nextArgs";
+const NEXT_ARGS_SCHEMA: &str = "NextArguments";
+const PUBLIC_KEY_ARG: &str = "publicKey";
+const PUBLIC_KEY_SCHEMA: &str = "PublicKey";
+const VIEW_ID_ARG: &str = "viewId";
+const VIEW_ID_SCHEMA: &str = "DocumentViewId";
```
adzialocha, line 19: Similarily to the other comment it could be nice to explain what these constants are for in a docstring. Also, is the naming here differently or shouldn't it also be "_SCALAR" instead of "_SCHEMA"?

aquadoggo/src/graphql/utils.rs
```diff
@@ -0,0 +1,61 @@
...
+///
+/// Panics when given a relation field value.
+pub fn gql_scalar(operation_value: &OperationValue) -> Value {
+    match operation_value {
+        OperationValue::Boolean(value) => value.to_owned().into(),
+        OperationValue::Integer(value) => value.to_owned().into(),
+        OperationValue::Float(value) => value.to_owned().into(),
+        OperationValue::String(value) => value.to_owned().into(),
+        // Recursion not implemented yet.
+        _ => todo!(),
```
adzialocha, line 31: What is this about? 🤔

aquadoggo/src/replication/service.rs
```diff
@@ -19,21 +19,20 @@ use tokio::task;
...
-use crate::graphql::replication::client;
 use crate::manager::{ServiceReadySender, Shutdown};
 
 /// Replication service polling other nodes frequently to ask them about new entries from a defined
 /// set of authors and log ids.
 pub async fn replication_service(
     context: Context,
     shutdown: Shutdown,
-    tx: ServiceSender,
+    _tx: ServiceSender,
```
adzialocha, line 29: I'd rather remove everything related to replication for now then having dead code in the codebase. We can always look into the .git history to see how we can reuse code from the past.

aquadoggo/src/tests.rs
```diff
@@ -368,6 +368,7 @@ async fn query(
     );
 
     let response = post(client, &query_str).await;
+    print!("{:?}", response.errors);
```
adzialocha, line 364: Think this needs to be removed?

adzialocha (changes requested):
I can't believe how easy it is to understand this and at the same time it is super powerful. It's really nicely done! 

I've added some smaller comments in my review, some general remarks:

* You have a whole bunch of "TODO" markers in the code which partly need to be extended or discussed. My suggestion would be to remove them and rather create GitHub issues instead? Then we can continue discussion there.
* There is a lot of underscores 😆 Don't think we make use of that naming pattern so far, what was the intention here? Is this a sign of too much dead code which needs to be removed? In general I think its good to not keep dead code around at any point.

aquadoggo/src/graphql/schema_builders/document.rs
sandreae: Great point, I'll add some docs.

aquadoggo/src/graphql/schema_builders/document.rs
sandreae: Lolz, I nearly wrote "`fields` field", will change it to that.

aquadoggo/src/graphql/schema_builders/document.rs
sandreae: Yep, did this 👍

aquadoggo/src/graphql/utils.rs
sandreae: Ah yeah originally I thought I'd implement the relation logic here, but actually no... This can either just be a panic, or move the method content in place, I think it's only used in one place.

aquadoggo/src/replication/service.rs
sandreae: Sure thing, let's do it.

aquadoggo/src/tests.rs
sandreae: Yes!

aquadoggo/src/graphql/schema.rs
sandreae: Ah, it would be if we refer to a type which isn't registered I think, which we shouldn't be doing.

aquadoggo/src/graphql/schema_builders/document.rs
sandreae: I think it would look the same, not really important as the current flow is fine.

aquadoggo/src/graphql/utils.rs
sandreae: This method is only used for converting concrete values, not relations.

adzialocha:
Clippy fails because of #295, will merge now anyhow. The replication service refactoring / removal I'll do in a separate PR :+1: Thank you for this amazing work @sandreae!

aquadoggo/src/graphql/schema_builders/document.rs
```diff
@@ -0,0 +1,458 @@
...
+        .field(Field::new(
+            constants::FIELDS_FIELD,
+            TypeRef::named(document_fields_name),
+            move |ctx| {
+                FieldFuture::new(async move {
+                    // Here we just pass up the root query parameters to be used in the fields resolver
+                    let params = downcast_id_params(&ctx);
+                    Ok(Some(FieldValue::owned_any(params)))
+                })
+            },
```
cafca, line 33: This 😭😭😭 So much better

Issue #303, opened 2023-03-17
Introduce `identify` and `rendezvous` network protocols


mycognosist:
The next step in our networking efforts is to extend the peer discovery functionality of `aquadoggo`.

The `libp2p` rendezvous protocol ([spec](https://github.com/libp2p/specs/tree/master/rendezvous)) provides a discovery mechanism by allowing peers to register themselves with a server and query the addresses of other peers. The rendezvous protocol is often paired with the identify protocol ([spec](https://github.com/libp2p/specs/tree/master/identify)) to provide a means for rendezvous client nodes to learn their external IP address (necessary knowledge for registering with a rendezvous server).

We are aiming for a single node (`aquadoggo`) which can run as a rendezvous server or client, depending on CLI configuration.

A rendezvous client needs the address and peer ID of a rendezvous server in order to connect, register and query.

PR #304, opened 2023-03-17
Introduce `identify` and `rendezvous` network protocols

mycognosist:
Addresses https://github.com/p2panda/aquadoggo/issues/303

- [x] Move `NetworkBehaviour` into a separate module (`behaviour.rs`)
- [x] Add the `identify` network behaviour / protocol
  - [x] Manually add an external address when reported via remote peer
  - [x] Handle `identify::Event::Received` event
  - [x] Handle `identify::Event::Sent` event
  - [x] Handle `identify::Event::Error` event
- [x] Add the `rendezvous` network behaviour / protocol
  - [x] Handle `rendezvous::client::Event::Registered` event
  - [x] Handle `rendezvous::client::Event::Discovered` event
  - [x] Handle `rendezvous::client::Event::Failed` event
  - [x] Handle `rendezvous::server::Event::PeerRegistered` event
  - [x] Handle `rendezvous::server::Event::DiscoverServed` event
- [x] Add configuration options for rendezvous
- [x] Expose rendezvous configuration options via CLI
- [x] Replace `String` with `Multiaddr` for `--remote-node-addresses` and `--rendezvous-address` (results in validation by parser)
- [x] Replace `String` with `PeerId` for `--rendezvous-peer-id` (results in validation by parser)
- [x] Add `glyph` to list of authors
- [x] Update CLI `README` with latest usage instructions

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/network/config.rs
+++ b/aquadoggo/src/network/config.rs
@@ -5,6 +5,7 @@ use std::path::PathBuf;
 use anyhow::Result;
 use libp2p::identity::Keypair;
 use libp2p::swarm::ConnectionLimits;
+use libp2p::{Multiaddr, PeerId};
 use log::info;
 use serde::{Deserialize, Serialize};
 
@@ -75,7 +76,7 @@ pub struct NetworkConfiguration {
     pub quic_port: u16,
 
     /// The addresses of remote peers to replicate from.
-    pub remote_peers: Vec<String>,
+    pub remote_peers: Vec<Multiaddr>,
 
     /// Rendezvous client behaviour enabled.
     ///
@@ -88,10 +89,10 @@ pub struct NetworkConfiguration {
     pub rendezvous_server: bool,
 
     /// Address of a rendezvous server in the form of a multiaddress.
-    pub rendezvous_address: Option<String>,
+    pub rendezvous_address: Option<Multiaddr>,
 
     /// Peer ID of a rendezvous server.
-    pub rendezvous_peer_id: Option<String>,
+    pub rendezvous_peer_id: Option<PeerId>,
 }
 
 impl Default for NetworkConfiguration {
--- a/aquadoggo/src/network/config.rs
+++ b/aquadoggo/src/network/config.rs
@@ -14,7 +14,6 @@ use crate::network::identity::Identity;
 /// Key pair file name.
 const KEY_PAIR_FILE_NAME: &str = "private-key";
 
-// TODO: make the namespace dynamic (ie. can be user-assigned)
 /// The namespace used by the `identify` network behaviour.
 pub const NODE_NAMESPACE: &str = "aquadoggo";
 
--- a/aquadoggo/src/network/config.rs
+++ b/aquadoggo/src/network/config.rs
@@ -13,6 +13,10 @@ use crate::network::identity::Identity;
 /// Key pair file name.
 const KEY_PAIR_FILE_NAME: &str = "private-key";
 
+// TODO: make the namespace dynamic (ie. can be user-assigned)
+/// The namespace used by the `identify` network behaviour.
+pub const NODE_NAMESPACE: &str = "aquadoggo";
+
 /// Network config for the node.
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct NetworkConfiguration {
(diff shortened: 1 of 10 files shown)


aquadoggo/src/network/config.rs
```diff
@@ -13,6 +14,9 @@ use crate::network::identity::Identity;
 /// Key pair file name.
 const KEY_PAIR_FILE_NAME: &str = "private-key";
 
+/// The namespace used by the `identify` network behaviour.
+pub const NODE_NAMESPACE: &str = "aquadoggo";
```
mycognosist, line 18: We might want to make this user-definable in the future. The namespace value is used when registering a rendezvous client with a rendezvous server and querying for other registered peers.

aquadoggo/src/network/mod.rs
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+mod behaviour;
```
mycognosist, line 3: Moving the network behaviour into a module makes the service less busy.

aquadoggo/src/network/service.rs
```diff
@@ -103,13 +66,20 @@ pub async fn network_service(
     // Listen for incoming connection requests over the QUIC transport
     swarm.listen_on(quic_multiaddr)?;
 
-    // Dial the peer identified by the multi-address given in the `--remote-node-addresses` if given
-    if let Some(addr) = network_config.remote_peers.get(0) {
-        let remote: Multiaddr = addr.parse()?;
-        swarm.dial(remote)?;
+    // Dial each peer identified by the multi-address provided via `--remote-node-addresses` if given
+    for addr in network_config.remote_peers.clone() {
```
mycognosist, line 70: Ensure we dial all remote nodes and not just the first one in the list.

aquadoggo/src/network/service.rs
```diff
@@ -140,13 +110,34 @@ pub async fn network_service(
...
+                        && network_config.rendezvous_peer_id.unwrap() == peer_id =>
+                {
+                    debug!(
+                        "Connected to rendezvous point, discovering nodes in '{NODE_NAMESPACE}' namespace ..."
+                    );
+
+                    swarm
+                        .behaviour_mut()
+                        .rendezvous_client
+                        .as_mut()
```
mycognosist, line 124: This looks messy, unfortunately. We have to call `.as_mut()` because the `rendezvous_client` behaviour is behind a `Toggle` type.

aquadoggo/src/network/service.rs
```diff
@@ -140,13 +110,34 @@ pub async fn network_service(
                 } => {
                     debug!("ConnectionClosed: {peer_id} {endpoint:?} {num_established} {cause:?}")
                 }
+                SwarmEvent::ConnectionEstablished { peer_id, .. }
+                    if network_config.rendezvous_client
+                        && network_config.rendezvous_peer_id.unwrap() == peer_id =>
```
mycognosist, line 115: We should always be safe to unwrap the `rendezvous_peer_id` here because the CLI will ensure it's provided if `rendezvous_client` was set to `true`.

aquadoggo/src/network/service.rs
```diff
@@ -179,7 +170,108 @@ pub async fn network_service(
...
+                                let p2p_suffix = Protocol::P2p(*peer.as_ref());
+                                let address_with_p2p = if !address
+                                    .ends_with(&Multiaddr::empty().with(p2p_suffix.clone()))
+                                {
+                                    address.clone().with(p2p_suffix)
+                                } else {
+                                    address.clone()
+                                };
+
+                                swarm.dial(address_with_p2p).unwrap();
```
mycognosist, line 204: Dial every peer we learn about from the rendezvous server.

aquadoggo_cli/src/main.rs
```diff
@@ -25,7 +27,7 @@ struct Cli {
 
     /// URLs of remote nodes to replicate with.
     #[arg(short, long)]
-    remote_node_addresses: Vec<String>,
+    remote_node_addresses: Vec<Multiaddr>,
```
mycognosist, line 30: By using `Multiaddr` and `PeerId` (instead of `String`) we get validation of the input automatically :) No need to try and `.parse()` later on.

aquadoggo/src/network/service.rs
```diff
@@ -140,13 +110,34 @@ pub async fn network_service(
...
+                {
+                    debug!(
+                        "Connected to rendezvous point, discovering nodes in '{NODE_NAMESPACE}' namespace ..."
+                    );
+
+                    swarm
+                        .behaviour_mut()
+                        .rendezvous_client
+                        .as_mut()
+                        .unwrap()
```
adzialocha, line 125: What could happen here?

adzialocha (commented):
Looking all very good, mostly just small questions ☺️

aquadoggo/src/network/config.rs
mycognosist: Agreed, sharing a common identifier would be best. mDNS uses a 'service name' which is currently not user-configurable in the Rust implementation (see https://github.com/libp2p/rust-libp2p/issues/2898 for more detail).

adzialocha (approved):

aquadoggo/src/network/service.rs
mycognosist: Actually, I found a way to make this slightly cleaner by using `if let` to avoid the `is_enabled()` check and `unwrap()` calls.

sandreae (approved):
All looks really good to me, don't have any change requests!

:clap: :clap: :clap:

Issue #306, opened 2023-03-21
Introduce `relay` network protocol


mycognosist:
May also require `autonat`.

I've started working on this on the [`add_relay` branch](https://github.com/p2panda/aquadoggo/tree/add_relay).

**Links:**

- https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md
- https://docs.rs/libp2p/latest/libp2p/relay/index.html
- https://blog.ipfs.tech/2022-01-20-libp2p-hole-punching/

PR #314, opened 2023-03-28
Introduce `autonat` and `relay` network protocols

mycognosist:
Addresses https://github.com/p2panda/aquadoggo/issues/306

## 📋 Checklist


aquadoggo/Cargo.toml
```diff
@@ -19,7 +19,6 @@ edition = "2018"
 anyhow = "^1.0.62"
 async-graphql = { version = "^5.0.6", features = ["dynamic-schema"] }
 async-graphql-axum = "^5.0.6"
-async-recursion = "^1.0.0"
```
mycognosist, line 22: I discovered two unused dependencies using `cargo udeps`. Tests are passing so I'm guessing they are indeed not needed but please double-check.

aquadoggo/src/network/service.rs
```diff
@@ -270,6 +277,40 @@ pub async fn network_service(
...
+                }
+                SwarmEvent::Behaviour(BehaviourEvent::RelayClient(event)) => {
+                    debug!("Unhandled relay client event: {event:?}")
+                }
+                SwarmEvent::Behaviour(BehaviourEvent::Autonat(event)) => {
+                    match event {
+                        autonat::Event::StatusChanged { old, new } => {
+                            trace!("NAT status changed from {:?} to {:?}", old, new);
+
+                            if let Some(addr) = external_circuit_addr.clone() {
```
mycognosist, line 291: Previously I was only initiating this set of actions (add external circuit address and register with the rendezvous server) if the NAT status changed to Private. However, since public NAT status does not necessarily mean that direct connectivity is possible, I am initializing the relay client behaviour for any NAT status if `--relay-client` mode is enabled.

aquadoggo_cli/README.md
```diff
@@ -50,6 +50,21 @@ Options:
 
           eg. --rendezvous-peer-id "12D3KooWD3eckifWpRn9wQpMG9R9hX3sD158z7EqHWmweQAJU5SA"
 
+      --relay-client <RELAY_CLIENT>
```
mycognosist, line 53: I intend on simplifying the CLI in a future PR (see https://github.com/p2panda/aquadoggo/issues/319).

aquadoggo/src/network/config.rs
```diff
@@ -74,6 +74,19 @@ pub struct NetworkConfiguration {
     /// QUIC transport port.
     pub quic_port: u16,
 
+    /// Relay client behaviour enabled.
+    ///
+    /// Engage in relayed connections to remote peers.
```
sandreae, line 79: I'm not sure when a peer wouldn't want to have their messages relayed if direct connection is impossible. Is there a use case in mind for this being optional?

aquadoggo/src/network/service.rs
```diff
@@ -31,42 +29,47 @@ pub async fn network_service(
...
+    let quic_multiaddr =
+        format!("/ip4/0.0.0.0/udp/{}/quic-v1", network_config.quic_port).parse()?;
 
     // Listen for incoming connection requests over the QUIC transport
     swarm.listen_on(quic_multiaddr)?;
 
-    // Dial each peer identified by the multi-address provided via `--remote-node-addresses` if given
+    let mut external_circuit_addr = None;
+
+    // Construct circuit relay addresses and listen on relayed address
```
sandreae, line 51: Sooo, this is the dedicated "relay message listener"?

aquadoggo/src/network/service.rs
sandreae: Does a nat status change mean our reachable address has changed?

aquadoggo/src/network/swarm.rs
```diff
@@ -0,0 +1,42 @@
...
+use libp2p::swarm::SwarmBuilder;
+use libp2p::PeerId;
+use libp2p::Swarm;
+use log::info;
+
+use crate::network::behaviour::Behaviour;
+use crate::network::transport;
+use crate::network::NetworkConfiguration;
+
+pub async fn build_swarm(
```
sandreae, line 16: Nice refactoring :clap:

aquadoggo/src/network/transport.rs
```diff
@@ -0,0 +1,56 @@
...
+        let relay_client = Some(relay_client);
+
+        // Add encryption and multiplexing to the relay transport
+        let relay_transport = relay_transport
+            .upgrade(Version::V1)
+            .authenticate(NoiseAuthenticated::xx(key_pair).unwrap())
+            .multiplex(YamuxConfig::default());
+
+        // The appropriate transport will be matched and utilised for connections (ie.
+        // relayed-connections will utilise the relay transport while all other incoming
```
sandreae, line 36: Does this mean relayed connections don't use QUIC (did we already discuss this, i feel like it was supported already...)?

sandreae (commented):
Hey, looks really great! I had a few questions to help my understanding but no concrete change requests.

aquadoggo/src/network/service.rs
mycognosist: > Does a nat status change mean our reachable address has changed?

Not quite. A NAT status change occurs when we receive data from a peer which tells us that our presumed public address is either reachable or not reachable. So the starting state is `Unknown`; this shifts to `Public` or `Private` based on insight passed to us by a peer (as part of the `autonat` protocol).

I was originally only executing the relay client-related code if / when the status changed to `Private` (ie. "oh, we're not publically accessible, we better register our circuit-relayed address so other peers can reach us via relayed connections"). However, I noticed that a `Public` NAT status does not always mean that a direct connection is possible (this is the case on my laptop, for example). So then I decided to simply execute the relay client-related code after the first status change, as long as relay client mode is active.

So, technically, the registration code-block could be moved elsewhere.

aquadoggo/src/network/config.rs
mycognosist: If the `autonat` protocol gave us an accurate and precise indication of reachability then I'd say that `relay_client` would not need to be optional. However, it is currently quite a useful "switch" to ensure that we only listen on a relayed address and register that address with a rendezvous server if we're in relay client mode. Does that make sense? Could definitely change in the future as `autonat` improves.

aquadoggo/src/network/service.rs
mycognosist: That's correct. It's our external relayed address (includes the address and peer ID of the relay server).

aquadoggo/src/network/service.rs
sandreae: Thanks for clarifying!

aquadoggo/src/network/transport.rs
mycognosist: https://github.com/libp2p/rust-libp2p/discussions/3766

mycognosist: As confirmed by Thomas, the relayed connections will use QUIC in our case. I'm going to quickly alter some of the variable naming to make this clearer in the code.

mycognosist:
Alrighty @sandreae , I reckon this should be good for a final check and mergeski.

aquadoggo/src/network/transport.rs
```diff
@@ -0,0 +1,57 @@
...
+    let (transport, relay_client) = if relay_client_enabled {
+        // Generate a relay transport and client behaviour if relay client mode is enabled
+        let (relay_transport, relay_client) = relay::client::new(key_pair.public().to_peer_id());
+        let relay_client = Some(relay_client);
+
+        // Add encryption and multiplexing to the relay transport
+        let relay_transport = relay_transport
+            .upgrade(Version::V1)
+            .authenticate(NoiseAuthenticated::xx(key_pair).unwrap())
+            .multiplex(YamuxConfig::default());
```
adzialocha, line 34: Looking at https://docs.rs/libp2p/latest/libp2p/relay/client/struct.Transport.html I understand that this whole block can be removed now, right?

mycognosist: Incorrect. The relayed connection still needs to be secured and multiplexed.

Take a look half-way through my comment here (where I write: "And the second suggestion (do not make the relay transport optional) results in a "mismatched types" error:"):

https://github.com/libp2p/rust-libp2p/discussions/3653#discussioncomment-5391803

adzialocha: Ah, thank you! I'm still not sure if I understand what is happening here though. How is the relay transport encrypted and multiplexed if it is not used in the end?

I read the comment in https://github.com/libp2p/rust-libp2p/discussions/3766:

> Relay itself isn't a transport. A relayed connection will always run on the connection that you establish to the relay

And that makes me wonder why at all this needs to be set then if all is handled by QUIC in the end?

adzialocha: Or in other words: Are we just setting these things to satisfy the API but actually relay transport is _not_ encrypted and multiplexed in these lines of code?

mycognosist: @sandreae has asked about this : https://github.com/libp2p/rust-libp2p/discussions/3766#discussioncomment-5595346

sandreae: This is an ongoing convo with the rust-libp2p devs, so let's not hold this PR up anymore :+1:

PR #318, opened 2023-03-28
Custom validation for GraphQL `Scalar` types

sandreae:
This PR implements custom validation on GraphQL scalar types. This means that any invalid errors passed into the api as arguments are internally validated and if they fail, errors are returned to the user with context information about which argument in the query failed and what value type was expected.

However, a side effect of this change is that in some cases (well, two i think) the actual validation errors returned from attempting to parse the value into it's expected type are lost. 

So there is a slight trade off involved here. In some places specific validation failure errors are lost, but in other places we gain useful context about which passed value was invalid.

See the changes for examples of the differences.

## 📋 Checklist


adzialocha:
> However, a side effect of this change is that in some cases (well, two i think) the actual validation errors returned from attempting to parse the value into it's expected type are lost.

> So there is a slight trade off involved here. In some places specific validation failure errors are lost, but in other places we gain useful context about which passed value was invalid.

Do we know why?

PR #339, opened 2023-04-25
Make `value` on `QueryRow` an option

sandreae:
## 📋 Checklist


aquadoggo/src/db/stores/query.rs
```diff
@@ -1571,6 +1573,61 @@ mod tests {
         });
     }
 
+    #[rstest]
+    fn empty_pinned_relation_list(key_pair: KeyPair) {
```
sandreae, line 1577: @adzialocha do you mind checking I've composed this test correctly? I'm expecting it to fail in the same way that the graphql one does, but it passes. Maybe I've misunderstanding how list are queried.

What I'm testing for is querying an application field on a pinned list field but when there are no item in the selected list.

adzialocha: Looks all good :+1: It makes a JOIN of the relation list with the documents inside of it. Since the relation list is empty, there's nothing to join, so all good :-)

sandreae: Thanks! I'll keep digging as to why the graphql fails then.

sandreae: Would you expect the two queries outlined in these tests (graphql and query store) to make the same SQL query?

PR #343, opened 2023-04-26
Move GraphQL types into separate modules

adzialocha:
This PR restructures the `graphql::types` module.

* Separate GraphQL Objects, Input Values and Responses into new modules
* Convert "constructors" to simple "build" methods
* Move resolvers into separate module, make them simple methods as well
* Rename `DocumentValue` to `Resolved` and implement `downcast` method on it
* Move `PaginationData` to `stores::query` as it is not related to GraphQL
* Rename `all_documents` query to `collection`

## 📋 Checklist

- [ ] ~~Link this PR to any issues it closes~~


adzialocha:
During development of the low-level query backend I stumbled always over not finding the methods which resolve certain parts of the GraphQL tree, so I thought it could be cool to move it all to a `resolvers` module.

There are more resolvers of course (for example for the `totalCount` field etc.) which are not in there, but I think that might be a bit too much for now? Most of the resolvers are also "only" there to downcast values.

sandreae (approved):
Yeh, looks good to me. Actually, my first design of these module builders/reducers was in a functional style like this, but I changed it along the way cos I thought it gave a familiar api between the dynamic and static types. I didn't really think it was a big success though. I like this :+1:

sandreae:
> There are more resolvers of course (for example for the totalCount field etc.) which are not in there, but I think that might be a bit too much for now? Most of the resolvers are also "only" there to downcast values.

Yeh, agree it's fine like this now.

PR #347, opened 2023-04-26
Fix SQL ordering to correctly assemble resulting rows to documents

sandreae:
## 📋 Checklist


sandreae:
Making the default ordering of query rows be by document id seems to have fixed the bug this PR is exploring. Some other tests fail now, but I assume their expected results just need updating. Not 100% sure the the original issue is taken care of though so will wait for second opinion before progressing.

aquadoggo/src/graphql/queries/collection.rs
```diff
@@ -386,4 +386,52 @@ mod tests {
             assert!(response.is_ok());
         });
     }
+
+    #[rstest]
+    fn a_funny_bug_which_needs_squishing(key_pair: KeyPair) {
```
adzialocha, line 391: Squish the bug 🐑

adzialocha:
I'll have a look at this, my gut feeling tells me that we want to sort by document view id _at the end_ of the ordering hierarchy instead

adzialocha:
> I'll have a look at this, my gut feeling tells me that we want to _always_ sort by document view id and _at the end_ of the ordering hierarchy instead, except of when it was selected manually.

I did this now and the test passes. I had something like this implemented before, it went lost somewhere during the journey 😆 

The addition of yours to set the default to "document id" is important, as it is part of the stack, but I think not part of this issue, let's add it in in another PR: https://github.com/p2panda/aquadoggo/issues/351

Now the proptests are failing, but I'm not sure if I understand yet how.

sandreae:
> I did this now and the test passes. I had something like this implemented before, it went lost somewhere during the journey laughing

Ah great, thanks.

sandreae:
> Now the proptests are failing, but I'm not sure if I understand yet how.

:thinking:

sandreae:
Wow, added some logging to figure out what's up with the prop tests, and ended up with this:

```rust
println!("Current documents len: {}", current_documents_len);
println!("Expected documents len: {}", expected_remaining_documents);
assert_eq!(
    current_documents_len, expected_remaining_documents,
    "Number of returned documents doesn't match expected remaining"
);
```

and the logging:

```
Current documents len: 2
Expected documents len: 2
thread 'proptests::tests::test_query' panicked at 'Test failed: assertion failed: `(left == right)`
  left: `1`,
 right: `2`: Number of returned documents doesn't match expected remaining;
```

sandreae:
Shame, these proptests have been really useful, but I am not debugging #%#$ like this.....

adzialocha:
> Wow, added some logging to figure out what's up with the prop tests, and ended up with this:
> 
> ```rust
> println!("Current documents len: {}", current_documents_len);
> println!("Expected documents len: {}", expected_remaining_documents);
> assert_eq!(
>     current_documents_len, expected_remaining_documents,
>     "Number of returned documents doesn't match expected remaining"
> );
> ```
> 
> and the logging:
> 
> ```
> Current documents len: 2
> Expected documents len: 2
> thread 'proptests::tests::test_query' panicked at 'Test failed: assertion failed: `(left == right)`
>   left: `1`,
>  right: `2`: Number of returned documents doesn't match expected remaining;
> ```

Wow, WTF?

adzialocha:
> I know right?? I almost threw my laptop across the room when I narrowed it down to this....
> 
> It happens every single run, different values, but same spot.

🤯 🤯 🤯

sandreae:
I removed the proptests for now, logged that issue on the proptests repo so we'll see if anything comes back about that.

PR #355, opened 2023-04-28
Troubleshoot issue in `proptest` tests

sandreae:
https://github.com/proptest-rs/proptest/issues/319

## 📋 Checklist


sandreae:
Turned out this was not a bug, but rather that the final log for the proptest result is printed out-of-order, so in this case it was rather confusing.

Issue #360, opened 2023-05-02
Bring back query proptests and move them behind a feature flag


adzialocha:
The CI should enable that feature flag

PR #362, opened 2023-05-08
Property tests for GraphQL query API

sandreae:
Introduce property tests which run against high-level GraphQL query api. This is done using the `proptest` library which works by defining strategies for generating test data. When failing tests are found the strategies perform "shrinking" which reduces the input values to their simplest form which still triggers the failure. The inputs for failed tests are stored in a `.regressions` file and the tests are run against these values first every time.

The following strategies are implemented:
- generate an AST representing schema with (nested) relations
  - target depth, fields and number of nodes can be configured
- generate collections of documents from schema AST
- generate query filters based on schema AST

The following tests are performed:
- pagination tests where the returned results are validated against expected values and pagination loops until completion
  - this is performed against a query for only meta data, and a query for only application fields
- perform a query which includes filter arguments for meta and application fields on the root query as well as any list fields
  - this does not check the returned collection for correctness, only tests that the queries do not error

Closes https://github.com/p2panda/aquadoggo/issues/360

## 📋 Checklist


.github/workflows/tests.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --manifest-path ${{ env.cargo_manifest }}
+          args: --manifest-path ${{ env.cargo_manifest }} --features=proptests
```
adzialocha, line 48: Let's move the prop test flow into an own action by namespacing the test :+1:

adzialocha (changes requested):

PR #382, opened 2023-05-18
Place GraphQL and networking modules behind feature flags for speedy dev compilation

sandreae:
## 📋 Checklist

### DIFF
--- a/aquadoggo/Cargo.toml
+++ b/aquadoggo/Cargo.toml
@@ -17,24 +17,27 @@ edition = "2018"
 
 [features]
 proptests = []
+graphql = ["dep:async-graphql", "dep:async-graphql-axum", "dep:axum", "dep:dynamic-graphql", "dep:http", "dep:tower-http"]
 
 [dependencies]
 anyhow = "1.0.62"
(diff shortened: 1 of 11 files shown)


adzialocha:
Hui! Is it getting too slow? How long are your compilation times?

I would move the features into the default feature set as we probably always want them, still one can opt out by setting the default feature flag to false.

sandreae:
> I would move the features into the default feature set as we probably always want them, still one can opt out by setting the default feature flag to false.

Yes, this is the plan.

sandreae:
> Hui! Is it getting too slow? How long are your compilation times?

The thing that's taking a while for me is `cargo check`.

If I run:
`cargo clean`
`cargo check --all-features --profile=test`

I get:
`Finished test [unoptimized + debuginfo] target(s) in 5m 34s`

Interested to see what you get here on your super computer ;-p

adzialocha:
I've closed this for now as I can't see any drastic regressions in complication time on my machines (including an old laptop). I like the feature flag idea though but that's maybe another PR?

PR #387, opened 2023-05-24
Integrate replication manager with networking stack

adzialocha:
This PR introduces a new replication service which represents the replication protocol layer on top of the networking stack. It uses the service bus to communicate with the networking service, gets informed about established and closed connection with peers and new incoming replication messages.

The replication service keeps track of all currently available peers, can initiate new replication sessions with a simple scheduler logic and maintains the sync manager which is the state machine dealing with the actual replication protocol logic.

Closes: https://github.com/p2panda/aquadoggo/issues/373 and https://github.com/p2panda/aquadoggo/issues/396

## 📋 Checklist


aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,274 @@
...
+        // Remove peer from our connections table
+        if self.peers.remove(&peer_id).is_none() {
+            warn!("Tried to remove unknown connection");
+        }
+
+        self.update_sessions().await;
+    }
+
+    async fn on_replication_message(&mut self, peer_id: PeerId, message: SyncMessage) {
+        match self.sync_manager.handle_message(&peer_id, &message).await {
```
adzialocha, line 142: Probably want to make the `IngestError` transparent so we understand better why arriving entries do not get accepted:

```
#[error(transparent)]
```

sandreae: I was getting a `ValidationError::SeqNumDoesNotMatch(1, 2)` error when logging this, but it's gone away now since i made some changes. Not sure _exactly_ what was causing it, but making sure to send the system schema over first definitely helped.

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,275 @@
...
+        match self.peers.get_mut(&peer_id) {
+            Some(status) => {
+                status.failed_count += 1;
+            }
+            None => {
+                panic!("Tried to access unknown peer");
+            }
+        }
+
+        // @TODO: SyncManager should remove session internally on critical errors
```
sandreae, line 190: Which cases does this message refer to @adzialocha ?

aquadoggo/src/replication/strategies/naive.rs
```diff
@@ -48,7 +58,10 @@ impl NaiveStrategy {
         let mut messages = Vec::new();
 
         let local_log_heights = self.local_log_heights(store).await;
-        let remote_needs = diff_log_heights(&local_log_heights, remote_log_heights);
+        let remote_needs = diff_log_heights(
+            &local_log_heights.to_owned().into_iter().collect(),
```
sandreae, line 62: This was the cause of https://github.com/p2panda/aquadoggo/issues/392 .... :sob:

aquadoggo/src/replication/service.rs
adzialocha: We haven't specified that yet but I assume all "critical" errors should terminate replication. I guess it also depends how strict we want to be.

aquadoggo/src/db/stores/document.rs
```diff
@@ -459,6 +459,7 @@ async fn insert_document_view(
             )
         VALUES
             ($1, $2, $3)
+        ON CONFLICT(document_view_id) DO NOTHING -- @TODO: temp fix for double document view insertions: https://github.com/p2panda/aquadoggo/issues/398
```
sandreae, line 462: Down and dirty fix for the materialization task queue duplication. (although, tbh, I don't see much wrong with solving it like this....)

aquadoggo/src/db/stores/entry.rs
```diff
@@ -336,7 +336,9 @@ impl SqlStore {
             WHERE
                 logs.schema = $1
             GROUP BY
-                entries.public_key, entries.log_id
+                entries.public_key, CAST(entries.log_id AS NUMERIC)
+            ORDER BY
+                entries.public_key, CAST(entries.log_id AS NUMERIC)
```
sandreae, line 341: Just ordering log heights cos it's of course nicer.

aquadoggo/src/network/replication/behaviour.rs
```diff
@@ -78,21 +104,49 @@ impl NetworkBehaviour for Behaviour {
...
             }
         }
     }
 
     fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>) {
         match event {
-            FromSwarm::ConnectionEstablished(_)
-            | FromSwarm::ConnectionClosed(_)
-            | FromSwarm::AddressChange(_)
+            FromSwarm::ConnectionClosed(ConnectionClosed {
```
sandreae, line 119: This is where we pass events up to the swarm which will then be caught and passed onto the service bus. The reason I did it here as opposed to directly on the swarm events is because we don't have access to `ConnectionId` there, which we want to pass onto the `ConnectionManager`.

aquadoggo/src/network/replication/handler.rs
```diff
@@ -208,7 +212,8 @@ impl ConnectionHandler for Handler {
                                 HandlerOutEvent::Message(message),
                             ));
                         }
-                        Poll::Ready(Some(Err(_))) => {
+                        Poll::Ready(Some(Err(err))) => {
+                            warn!("{err:#?}");
```
sandreae, line 216: Just logging these deserialization errors in handler right now.

aquadoggo/src/network/service.rs
```diff
@@ -259,68 +419,48 @@ pub async fn network_service(
...
                 }
-                SwarmEvent::Behaviour(BehaviourEvent::Limits(event)) => {
-                    debug!("Unhandled connection limit event: {event:?}")
+                replication::Event::ConnectionClosed(peer_id, connection_id) => {
+                    // Inform other services about closed connection
+                    self.send_service_message(ServiceMessage::ConnectionClosed(
+                        peer_id,
+                        connection_id,
+                    ));
                 }
```
sandreae, line 457: This is where we catch the events triggered in the replication behaviour and pass them onto the service bus.

aquadoggo/src/replication/errors.rs
```diff
@@ -49,3 +52,28 @@ pub enum TargetSetError {
...
+    )]
+    InboundExistingTargetSet(TargetSet),
+
+    #[error(
+        "Tried to initialise duplicate outbound replication session for existing target set {0:?}"
+    )]
+    OutboundExistingTargetSet(TargetSet),
+
+    #[error("Tried to initialise duplicate outbound replication session with id {0}")]
+    Outbound(u64),
```
sandreae, line 78: A load of errors relating to our de-duplication logic in sync manager..... Not super happy with this yet, but it's nice to have custom error types for these cases.

aquadoggo/src/replication/manager.rs
```diff
@@ -176,47 +200,131 @@ where
...
+            // different target set?
+        }
+
+        Ok(SyncResult {
+            messages: all_messages,
+            is_done: false,
+        })
+    }
+
+    async fn handle_duplicate_target_set(
```
sandreae, line 272: Added this method for handling session requests with duplicate `TargetSet`s in the same way as we do for those with duplicate session ids.

sandreae: Sure these two methods could be refactored to avoid some code duplication too....

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,398 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}{:?}",
+            self.0,
+            // If we don't pass a connection id then we default to `0`
+            self.1.unwrap_or(ConnectionId::new_unchecked(0))
+        )
+    }
+}
```
sandreae, line 47: New struct used for identifying sessions.

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,398 @@
...
+    // Run a connection manager which deals with the replication logic
+    let manager = ConnectionManager::new(
+        &context.schema_provider,
+        &context.store,
+        &tx,
+        local_peer_id,
+        target_set,
+    );
+
+    let scheduler_handle = task::spawn(ConnectionManager::start_scheduler(tx.clone()));
```
sandreae, line 83: In keeping with our naive replication, we get some naive scheduling too ;-p

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,398 @@
...
+    connections: Vec<ConnectionId>,
+}
+
+struct ConnectionManager {
+    connections: HashMap<PeerId, PeerConnections>,
+    sync_manager: SyncManager<PeerConnectionId>,
+    tx: ServiceSender,
+    rx: Receiver<ServiceMessage>,
+    target_set: TargetSet,
+}
```
sandreae, line 131: I extended the map of peers to include all connections to each peer (copied from libp2p gossip protocol implementation).

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,398 @@
...
+                self.on_connection_established(peer_id, connection_id).await;
+            }
+            ServiceMessage::ConnectionClosed(peer_id, connection_id) => {
+                self.on_connection_closed(peer_id, connection_id).await;
+            }
+            ServiceMessage::ReceivedReplicationMessage(peer_id, connection_id, message) => {
+                self.on_replication_message(peer_id, message, connection_id)
+                    .await;
+            }
+            ServiceMessage::InitiateReplication => {
```
sandreae, line 302: This comes from the scheduler and is now the only moment replication gets triggered.

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    fn send_service_message(&self, message: ServiceMessage) {
+        if self.tx.send(message).is_err() {
+            // Silently fail here as we don't care if the message was received at this
+            // point
+        }
+    }
+
+    async fn update_sessions(&mut self) {
```
sandreae, line 316: Need to revisit this logic to check it's still doing what we expect (start a session for all connections we know about)

aquadoggo/src/bus.rs
```diff
@@ -12,4 +15,19 @@ pub type ServiceSender = Sender<ServiceMessage>;
 pub enum ServiceMessage {
     /// A new operation arrived at the node.
     NewOperation(OperationId),
+
+    /// Message from scheduler to initiate replication on available peers.
+    InitiateReplication,
```
adzialocha, line 20: Why is this a bus message? Is it needed across services?

aquadoggo/src/db/stores/document.rs
adzialocha, line 462: Interesting, could that already be a fix? What are the implications of this change?

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,398 @@
...
+                Err(err) => {
+                    panic!("Service bus subscriber failed: {}", err);
+                }
+            }
+        }
+    }
+
+    pub async fn start_scheduler(tx: ServiceSender) {
+        loop {
+            sleep(Duration::from_secs(5)).await;
```
sandreae, line 391: :100:

aquadoggo/src/replication/service.rs
```diff
@@ -0,0 +1,398 @@
...
+        Self {
+            peer_id: *peer_id,
+            successful_count: 0,
+            failed_count: 0,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct PeerConnections {
```
adzialocha, line 120: We could probably find a way to merge these structs `PeerStatus` and `PeerConnections` now, or at least have the `PeerId` field on the primary one.

aquadoggo/src/replication/target_set.rs
```diff
@@ -52,13 +66,45 @@ impl Validate for TargetSet {
...
+                        return Err(TargetSetError::UnsortedSchemaIds);
+                    }
+                    // If current and previous are both system schema then compare them.
+                    _ if initial_system_schema => {
+                        if prev >= schema_id {
+                            return Err(TargetSetError::UnsortedSchemaIds);
+                        }
+                    }
+                    _ => panic!(),
+                };
```
sandreae, line 107: This is where the validation was failing when deserializing `TargetSet`r

aquadoggo/src/replication/service.rs
adzialocha, line 83: Wondering if that scheduler can be an async stream which wakes up the runtime every x seconds, there's some nice rust patterns for it. I can look something up!

sandreae (commented):
Just some connents to help you grok the changes I made in the last few days @adzialocha .

aquadoggo/src/replication/target_set.rs
```diff
@@ -52,13 +66,45 @@ impl Validate for TargetSet {
         };
 
         let mut prev_schema_id: Option<&SchemaId> = None;
+        let mut initial_system_schema = true;
+
+        // We need to validate that:
+        // - if system schema are included they are first in the list and ordered alphabetically
+        // - any following application schema are also ordered alphabetically
+        for (index, schema_id) in self.0.iter().enumerate() {
```
adzialocha, line 74: Do we need any tests checking that new validation logic?

aquadoggo/src/network/replication/behaviour.rs
sandreae: This goes some way to solving https://github.com/p2panda/aquadoggo/issues/399 as when a connection suddenly closes it eventually times out and a `ConnectionClosed` event is sent. We may also need to work on the `KeepAlive` logic in the handler, we could give it a `KeepAlive::Until(x)` which might help libp2p tidy up stale connections.

aquadoggo/src/db/stores/document.rs
adzialocha: Yeah I'm also curious to hear what you think about it, what implications of such change do you see?

adzialocha:
> Just some connents to help you grok the changes I made in the last few days @adzialocha .

Great, thanks! That's very helpful and clear. I think I added some comments while you posted yours, but I think they still are up to date.

aquadoggo/src/bus.rs
sandreae: > Why is this a bus message? Is it needed across services?

I didn't have that in mind, I just needed to get messages from the network service -> replication service and thought this was the way to do it.... Is there a better way?

aquadoggo/src/db/stores/document.rs
sandreae: Yeh, I like it, can't see why not ;-p

aquadoggo/src/bus.rs
adzialocha: > network service -> replication service

I only see this being used in Replication Service

aquadoggo/src/db/stores/document.rs
adzialocha: > Yeh, I like it, can't see why not ;-p

Haha, rock'n'roll.

But also, I'd like really to understand what this change means 😆

adzialocha: > I think the only implications are in terms of potential errors being hidden away, but in this case, I don't think there are bad pitfalls 

Can you give some examples?

aquadoggo/src/replication/errors.rs
adzialocha: I like it, that might make it easier to debug in case something goes wrong

aquadoggo/src/bus.rs
sandreae: Oh yeh, sorry, I was thinking the comment included the other new variants too! Yeh, this one is only used in the Replication Service :+1:

aquadoggo/src/replication/service.rs
adzialocha: Great! I was thinking about adding something like this as well, think its already a really good first version until we get "live mode".

aquadoggo/src/bus.rs
adzialocha: > I was thinking the comment included the other new variants too

Hehe yeah, the other "new variants" I actually added :-D

adzialocha: > Yeh, this one is only used in the Replication Service

Cool, I think using the service bus for this is too global, it can be either a channel again or even better an async stream which wakes up every x seconds (as I've written in another comment).

aquadoggo/src/db/stores/document.rs
sandreae: Well, like now, because of this restriction being in place, we know two tasks can exist in the queue which are both resolving the same document view. Guess this isn't a bug, but it tells us about a behaviour we might want to account for. 

The document view row will never change once inserted, so we shouldn't be operating on it again, and we shouldn't be inserting it again. Keeping it as it was before tells us if do either of these things by accident.

adzialocha: > The document view row will never change once inserted

If thats the case, Why are even two document views created? Maybe the issue is actually in one of the worker logics?

aquadoggo/src/network/replication/behaviour.rs
```diff
@@ -219,36 +287,63 @@ mod tests {
...
+                Event::ConnectionEstablished(peer_id, _) = swarm1.next_behaviour_event() => res1.push((peer_id, None)),
+                Event::ConnectionEstablished(peer_id, _) = swarm2.next_behaviour_event() => res2.push((peer_id, None)),
+            }
+        }
+
+        // And again add the next 2 behaviour events which occur in either swarms.
+        for _ in 0..2 {
+            tokio::select! {
+                Event::MessageReceived(peer_id, message, _) = swarm1.next_behaviour_event() => res1.push((peer_id, Some(message))),
+                Event::MessageReceived(peer_id, message, _) = swarm2.next_behaviour_event() => res2.push((peer_id, Some(message))),
```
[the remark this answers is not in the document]
adzialocha: This fails with "Peer id needs to be given" because of my change moving the `PeerId` to the `NetworkConfiguration` struct. I need to refactor this properly .. a quick change didn't do it.

aquadoggo/src/db/stores/document.rs
sandreae: Yeh, i think it's a concurrency issue in the task issuing logic, something like:

- document A has a relation to document B at view 1
- reduce task for document A is in the task queue
- dependency task for document B at view 1 is issued
== operation creating document B at view 1 arrives on the node ==
- reduce task for document B is issued

now two tasks will resolve document B to view 1 and insert it into the db.

adzialocha: Agreed that the SQL constraint is very strict here and likely not needed, but I think we should only have it when a) it gives us performance benefits (less checks required, better concurrency etc.) and b) we understand that there is not another bug hiding somewhere.

I need to read through the `reduce` and `dependency` worker code (as you said its been running when the SQL UNIQUE error came up). We might actually deal with a bug there or even worse with the de-duplication logic (though, less likely).

aquadoggo/src/replication/target_set.rs
sandreae: We sure do: https://github.com/p2panda/aquadoggo/pull/387/commits/c5373eff837ff15c36542480fe418eaa63b8913d

adzialocha: Amazing!

aquadoggo/src/db/stores/document.rs
sandreae: Yeh, very happy to figure out if we are confident the worker logic is correct before settling on anything here.

adzialocha: > Yeh, i think it's a concurrency issue in the task issuing logic, something like:
> 
>     * document A has a relation to document B at view 1
> 
>     * reduce task for document A is in the task queue
> 
>     * dependency task for document B at view 1 is issued
>       == operation creating document B at view 1 arrives on the node ==
> 
>     * reduce task for document B is issued
> 
> 
> now two tasks will resolve document B to view 1 and insert it into the db.

Aha, so the problem is that we have two concurrent reduce tasks which are considered _different_ since one of them addresses the document via its _id_, and the other via its _view id_?

- `reduce(<abcd ... document id>, None)`
- `reduce(None, <abcd ... document view id>)`

adzialocha: Need to think more about this, doesn't sound like an concurrency issue too me either right now (the case I describe above can also happen in series). Maybe the we need to have an upsert instead (either in the `reduce` task or on SQL level as this fix already does).

adzialocha: I'll mark it as "solved" here, since we will deal with this in a separate PR

aquadoggo/src/network/service.rs
```diff
@@ -79,177 +83,333 @@ pub async fn network_service(
...
+                        debug!("mDNS discovered a new peer: {peer_id}");
+
+                        if let Err(err) = self.swarm.dial(multiaddr) {
+                            warn!("Failed to dial: {}", err);
                         }
+
+                        // // Only dial the newly discovered peer if we're not already connected.
+                        // //
+                        // // @TODO: Is this even a thing? Trying to catch the case where two peers
+                        // // simultaneously discover and connect to each other.
```
adzialocha, line 260: Interesting! So this called `swarm.dial` twice?

aquadoggo/src/replication/service.rs
adzialocha: Probably the maximum needs to be set to 1 right now since we're always using the same target set for each peer? So any other second replication session would be considered invalid (since it already exists)?

Think we're never seeing that problem since our replication is too fast with that little data.

adzialocha: Also, fixed that filter logic issue:

```rust
let active_sessions: Vec<&Session> = sessions
    .iter()
    .filter(|session| !session.is_done())
    .collect();
```

.. where it only returned finished sessions before.

adzialocha: > Probably what we need here is to check if that target set is already currently being replicated with that peer. We want max. one session per target set per peer, independent of established locally or remotely.

I've implemented that now

aquadoggo/src/replication/manager.rs
```diff
@@ -452,11 +587,350 @@ mod tests {
...
+    //  PEER A                                 PEER B
+    //
+    //  SyncRequest(0, 0, ["A"])────────────────────►
+    //
+    //  ◄─────────────────── SyncRequest(0, 0, ["B"])
+    //
+    //  ========== PEER A REQUEST DROPPED ===========
+    //
+    //  0 Have([..]) ───────────────────────────────►
+    //  0 SyncDone(false) ──────────────────────────►
```
adzialocha, line 604: Think the diagrams need to look like that (no `SyncDone` being moved to the end). Our messages are FIFO-based, messages arrive in the order they were sent, one by one, nothing gets re-ordered. The diagram before suggested that the first `SyncDone` message gets ordered last, not sure if that was indended though.

sandreae: It was actually intentional to do it as it was, the reason being that I wanted to demonstrate the order messages were processed on each node, as well as when theyre sent. Think this is most useful for understanding the state changes. Peer B processes Peer A's `Have` message, generates and sends  a `Have` and `SyncDone`, _then_ processes the `SyncDone`.

adzialocha: Cool, I understand now!

aquadoggo/src/network/service.rs
```diff
@@ -79,177 +83,338 @@ pub async fn network_service(
...
+            }
+            SwarmEvent::NewListenAddr {
+                address,
+                listener_id: _,
+            } => {
+                info!("Listening on {address}");
+            }
+            SwarmEvent::OutgoingConnectionError { peer_id, error } => {
+                warn!("OutgoingConnectionError: {peer_id:?} {error:?}");
+                // self.send_service_message(ServiceMessage::ConnectionError(peer_id));
```
adzialocha, line 247: Dead code?

sandreae:
So with these last commits I am trying to handle the case where peer A and peer B have an existing connection (connection 1), then peer B disconnects and then reconnects to peer A quickly (connection 2). Until connection 1 times-out, peer A doesn't know there was a dropped connection, and so sync messages arriving from the new peer (on connection 2) may connect to not-yet-closed replication sessions.

My approach was to limit connections between peers to one inbound and one outbound connection, rejecting all other connections, and only triggering `PeerDisconnected` events when both of our know connections are dropped.

sandreae:
This seems to work, peers are blocked from re-establishing a connection when they suddenly disconnected, and when the original connection times-out then they can once again connect. For it to work "in-the-wild" we would need some re-dial logic as once a peer is blocked from connecting then they will never automatically try again.

sandreae:
All this threw up a something that confused me and sent me down a little libp2p rabbit hole... which is the fact that if you repeatedly disconnect and reconnect one peer, you will hit a "Max established connection limit reached" error, which was surprising as I expected all connections to be nicely closed when rejected. This must not be the case, and it would be nice to know what's going on there.

adzialocha:
> This must not be the case, and it would be nice to know what's going on there.

Your current code only disallows protocol upgrades for our custom behavior / handler when a limit was hit. So other behaviours will still allow this connection.

I'm also not sure yet if that blocking behavior is what we want, it assumes shared state between nodes (now the other node needs to know if it should re-dial).

sandreae:
Accidental closure 😅

adzialocha:
> > Your current code only disallows protocol upgrades for our custom behavior / handler when a limit was hit. So other behaviours will still allow this connection.
> 
> Ok, yeh, that's true. I hadn't fully appreciated that 🤔 so the other handlers will keep this connection alive, which explains the "Max establish connection limit" being hit. We could trigger a `disconnect_peer` on the swarm as well.

I'm saying that, but then checked the libp2p documentation and it says something different ("connection will be closed immediately") 🤔 - if we only have one - two connections per peer shared among all behaviours then your code should be right?!

Some days ago I used the `network_info` method of the Swarm, that was helpful to understand a little bit the current state of it.

sandreae:
> I'm also not sure yet if that blocking behavior is what we want, it assumes shared state between nodes (now the other node needs to know if it should re-dial).

I was thinking the node who blocked the request would redial, they know enough locally to do this. 

Did you have another approach in mind?

sandreae:
> I'm saying that, but then checked the libp2p documentation and it says something different ("connection will be closed immediately") 🤔 - if we only have one - two connections per peer shared among all behaviours then your code should be right?!

Hm, that is what I thought as well. As far as I understand it, there are 3 ways to close a connection, and it's not clear which is the best way in different cases, I might drop a question about this on the rust-libp2p repo. We have:

1) return an error from `handle_established_x_connection` in a behaviour
2) return `ConnectionHandler::Close(..)` from `poll` in a handler
3) return `KeepAlive::No` from `connection_keep_alive` in a handler

As I understand it, `1)` and `2)` should close the connection (regardless of other handlers) immediately, and `3)` does it in coordination with othe handlers.

sandreae:
> if we only have one - two connections per peer shared among all behaviours then your code should be right?!

It _could_ be that even when a node rejects/closes a connection themselves, there's still a time-out period where the connection hangs around. This would explain how connections build up even when it seems like they shouldn't be.

sandreae:
Rather relevant discussion: https://github.com/libp2p/rust-libp2p/discussions/3353

sandreae:
Also relevant "Deprecate `ConnectionHandlerEvent::Close`" https://github.com/libp2p/rust-libp2p/issues/3591

sandreae:
Interesting:

```
[2023-06-09T09:27:06Z WARN  aquadoggo::network::service] IncomingConnectionError: /ip4/0.0.0.0/udp/2022/quic-v1 /ip4/192.168.1.87/udp/2024/quic-v1 Denied { cause: ConnectionDenied { inner: Exceeded { limit: 8, kind: PendingIncoming } } }
[2023-06-09T09:27:06Z DEBUG aquadoggo::network::service] NetworkInfo {
        num_peers: 1,
        connection_counters: ConnectionCounters {
            limits: ConnectionLimits {
                max_pending_incoming: None,
                max_pending_outgoing: None,
                max_established_incoming: None,
                max_established_outgoing: None,
                max_established_per_peer: None,
                max_established_total: None,
            },
            pending_incoming: 1,
            pending_outgoing: 0,
            established_incoming: 1,
            established_outgoing: 0,
        },
    }
```

sandreae:
https://github.com/libp2p/rust-libp2p/discussions/4054

sandreae:
I changed the libp2p import to use the most up-to-date version from the git repo, this is basically `0.5.2` which contains a few breaking changes.

Interestingly `ConnectionId` is exposed to swarm in a few more events. I didn't notice any change in behaviour though.

sandreae:
> I changed the libp2p import to use the most up-to-date version from the git repo, this is basically `0.5.2` which contains a few breaking changes.
> 
> Interestingly `ConnectionId` is exposed to swarm in a few more events. I didn't notice any change in behaviour though.

We should revert this before continuing/merging, just wanted to test and see if it changed anything.

adzialocha:
> > I changed the libp2p import to use the most up-to-date version from the git repo, this is basically `0.5.2` which contains a few breaking changes.
> > Interestingly `ConnectionId` is exposed to swarm in a few more events. I didn't notice any change in behaviour though.
> 
> We should revert this before continuing/merging, just wanted to test and see if it changed anything.

I've reverted now without deleting that commit, then we can cherry-pick it later :+1:

aquadoggo/src/network/service.rs
```diff
@@ -79,177 +85,388 @@ pub async fn network_service(
...
+            rx: BroadcastStream::new(tx.subscribe()),
+            tx,
+            external_circuit_addr,
+            network_config,
+            shutdown_handler,
+        }
+    }
+
+    /// Close all connections actively.
+    pub async fn shutdown(&mut self) {
```
sandreae, line 144: This is so good, makes a huge difference to node UX :-]

sandreae (approved):
We finally made it :rofl: !!! Thanks for these final changes, it's really clear what's going on now and I'm observing very consistent behaviour when testing locally. Much better experience observing node behaviour now that the reaction to disconnects is immediate. 

This final approach to peers/connections should make it easy to transition to scheduling dialing peers from a behaviour, rather than in the sync manager.

I believe I'm still experiencing one issue which occurs when new schema are being materialized. It looks like a materialization task crashes and so the schema provider is never updated, if i restart the node, the provider is updated and replication can continue. Logs are here: https://laub.liebechaos.org/WQ3SzzmtSPyfESfLfNEgFA don't know if this is a concern of the current PR though, I can open an issue once I understand the problem better. 

I renamed `Naive` strategy to `LogHeight`. It's not perfect, but seems mean to call the current setup naive :smile:  

Would be good to figure out the cause of the issue (above) but apart from that I'm very happy with this PR now.

aquadoggo/src/network/service.rs
```diff
@@ -79,177 +85,388 @@ pub async fn network_service(
...
+
+            // ~~~~
+            // mDNS
+            // ~~~~
+            SwarmEvent::Behaviour(BehaviourEvent::Mdns(event)) => match event {
+                mdns::Event::Discovered(list) => {
+                    for (peer_id, multiaddr) in list {
+                        debug!("mDNS discovered a new peer: {peer_id}");
+
+                        if let Err(err) = self.swarm.dial(multiaddr) {
```
sandreae, line 319: I fixed this to dial one multiaddr at a time, this reduces duplicate outbound connections.

aquadoggo/src/network/peers/handler.rs
```diff
@@ -150,22 +220,36 @@ impl ConnectionHandler for Handler {
...
+            HandlerInEvent::CriticalError => {
+                self.critical_error = true;
+            }
         }
     }
 
     fn connection_keep_alive(&self) -> KeepAlive {
-        self.keep_alive
+        if self.critical_error {
+            return KeepAlive::No;
```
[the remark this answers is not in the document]
adzialocha: When _all_ behaviours flip KeepAlive to false then the connection gets removed (time out)

adzialocha:
> I believe I'm still experiencing one issue which occurs when new schema are being materialized. It looks like a materialization task crashes and so the schema provider is never updated, if i restart the node, the provider is updated and replication can continue. Logs are here: https://laub.liebechaos.org/WQ3SzzmtSPyfESfLfNEgFA don't know if this is a concern of the current PR though, I can open an issue once I understand the problem better.

Yes, I observed the same! It's caused by entries arriving in unexpected order (or at least some race condition). Also think it's unrelated to the networking stack

sandreae:
Wahoo!! MERGED!

adzialocha:
> Wahoo!! MERGED!

Juhuuu!!!

Issue #388, opened 2023-05-24
Handle key pair outside of network configuration


adzialocha:
Currently the keypair is too closely tied to libp2p.

1. The code should live outside of the `network` module and libp2p context
2. We want to use our own p2panda `KeyPair` type
3. Have a method to convert to the libp2p `Keypair` type to make it work with the libp2p universe
4. Nice code that generates and stores the key pair as soon as we start a node (in test or production environments), this is currently too closely tied to networking configuration (we want to use it ouside of networking contexts)

PR #408, opened 2023-06-19
Decouple p2panda's authentication data types from libp2p's

adzialocha:
## Separation of worlds

This separates the worlds of libp2p and p2panda a bit better. Before we would deal with libp2p's `Keypair` inside of our `Configuration` struct to handle and store secrets.

The only place where we "leak" libp2p into our other parts is the replication service, there we have still a notion of `PeerId`. We could wrap that around our own type to abstract it away, but at this point that seems too much.

## Making `KeyPair` more visible

This PR brings `KeyPair` into the picture by moving it out of `Configuration`, into its own `Context` field and on top into the `Node.start(key_pair, config)` method.

Closes https://github.com/p2panda/aquadoggo/issues/388

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/config.rs
+++ b/aquadoggo/src/config.rs
@@ -98,11 +98,6 @@ impl Configuration {
             }
         };
 
-        // Derive peer id from key pair
-        // @TODO: This needs refactoring: https://github.com/p2panda/aquadoggo/issues/388
-        let key_pair = NetworkConfiguration::load_or_generate_key_pair(config.base_path.clone())?;
-        config.network.set_peer_id(&key_pair.public());
-
         Ok(config)
     }
 }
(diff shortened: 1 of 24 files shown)


aquadoggo/src/config.rs
```diff
@@ -111,16 +106,9 @@ impl Configuration {
...
             ..Default::default()
-        };
-
-        // Generate a random key pair and just keep it in memory
-        // @TODO: This needs refactoring: https://github.com/p2panda/aquadoggo/issues/388
-        let key_pair: libp2p::identity::Keypair = crate::network::identity::Identity::new();
-        config.network.set_peer_id(&key_pair.public());
-
-        config
+        }
```
sandreae, line 112: I like the separation of identity from config :+1:

sandreae (approved):
Really nice refactor, I like how the separation allows us to imagine different networking solutions.

Issue #379, opened 2023-05-17
Querying a filtered collection breaks on updated documents


adzialocha:
The following query breaks aquadoggo after `eth_public_key` field got updated:

```gql
{
  profiles: all_profile_0020b1b79c08e8fc478f03bdb240ca1ae082c511b985e970dec9dffffdf0adf0ea76(
    filter: {
      eth_public_key: { eq: "0x7A4862B32Cfe424dDd30f232587C0cb2B032dc98" }
    }
  ) {
    documents {
      fields {
        eth_public_key
        username
      }
    }
  }
}

```

.. with the following error:

```
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', aquadoggo/src/graphql/resolvers.rs:134:30
```

The query _works_ when removing _either_ the `filter` _or_ the `eth_public_key` field:

- `filter` + `eth_public_key` = 💥 
- `eth_public_key` only = ✅ 
- `filter` only = ✅ 

Probably related: https://github.com/p2panda/aquadoggo/issues/356

adzialocha:
> I added one test which passes: https://github.com/p2panda/aquadoggo/commit/8cd21bdf2a7e012785a3b3429ff8343588dbbfe6
> 
> Maybe we need to add filtering/ordering or only update one of the documents in the collection.

Interesting, I'll share later which query failed. Might also be a nested relation 🤔

adzialocha:
Just noticed that your test doesn't touch the GraphQL bit but it failed in the resolvers

sandreae:
> Just noticed that your test doesn't touch the GraphQL bit, but the case we looked at failed in the resolvers logic, maybe the issue is there?

Yeh, should check there as well. There's not a lot (in terms of query logic) going on in the GraphQL though, so my instinct is that we can observe it here. Might be wrong about that of course.

adzialocha:
I've added more information now, we're getting closer:

> Interesting, I'll share later which query failed. Might also be a nested relation thinking

Don't think that's the case, since this query does not contain any nesting (the other one did).

> Maybe we need to add filtering/ordering or only update one of the documents in the collection.

Definitely breaking when adding filtering, don't think its about ordering or that it's only one document.

PR #409, opened 2023-06-19
Fix issue where filtered queries fail on updated fields

adzialocha:
Closes https://github.com/p2panda/aquadoggo/issues/379

## 📋 Checklist


adzialocha:
The problem is in this sub-select SQL query:

```
AND EXISTS (
    SELECT
        operation_fields_v1.value
    FROM
        operation_fields_v1
    WHERE
        operation_fields_v1.name = 'username'
        AND
            operation_fields_v1.value = $1
        AND
            operation_fields_v1.operation_id = document_view_fields.operation_id
)
```

It filters all fields out which do not fulfill this requirement (username = $1), even when they belong to the same document ..

adzialocha:
> So it is filtering out the other document fields, which we actually might want (depending on the selected fields)?

Exactly! The problem was that we just filtered document fields, but actually we wanted to filter document views which contain that document field. I extended the sub query now with a JOIN to reflect this.

sandreae (approved):

Issue #398, opened 2023-05-31
Double insertion of same view into `document_views` table during materialization


sandreae:
In the situation where the materializer task queue contains both:
- a `dependency` task for a document containing a relation to document `a` at view `3915a` 
- a `reduce` task for document `a` which resolves to view `3915a`

Then as both of these tasks are resolving and storing a view into the store then there will be a critical error:

```
Critical error in worker reduce with task <QueueItem 1 w. <TaskInput 00209f6b09271e1a71ae118d1db23247e3f9365a067d05a7c600f6294efe2853915a/->>: A fatal error occured in DocumentStore: error returned from database: (code: 1555) UNIQUE constraint failed: document_views.document_view_id
```

adzialocha:
The task manager deduplicates only within workers but not across them, this needs probably changing, I'll think about it a bit more but can't see any issues yet.

sandreae:
Temp fix: https://github.com/p2panda/aquadoggo/pull/387/files#diff-641af54a1fa4a1a4ab70451a0d3ae17354778caf603b3e20095101be68e45db7R462

adzialocha:
Related discussion: https://github.com/p2panda/aquadoggo/pull/387#discussion_r1212334388

sandreae:
Note: if we're happy to allow this double insertion, should it be solved in the SQL or in code?

PR #410, opened 2023-06-20
Handle duplicate document view insertions

sandreae:
Document views have a `UNIQUE` constraint on their `document_view_id` which means if double insertion of the same document view occurs then a fatal database error is returned. Although this is the expected behaviour, and there is no reason to want to update a document view (once inserted, no values on a document view should change), it can happen that a document is inserted twice because of two materializer tasks working towards the same document view (see issue and discussion here: https://github.com/p2panda/aquadoggo/issues/398).

This PR retains the `UNIQUE` restraint on document view id, and ~changes the behaviour of `insert_document_view` on `SqlStore`~ checks if a document view has already been materialized in the reduce task before performing any insertion.

Closes https://github.com/p2panda/aquadoggo/issues/398

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/db/stores/document.rs
+++ b/aquadoggo/src/db/stores/document.rs
@@ -30,6 +30,7 @@
 //! retained, we use a system of "pinned relations" to identify and materialise only views we
 //! explicitly wish to keep.
 use async_trait::async_trait;
+use log::debug;
 use p2panda_rs::document::traits::AsDocument;
 use p2panda_rs::document::{Document, DocumentId, DocumentView, DocumentViewId};
 use p2panda_rs::schema::SchemaId;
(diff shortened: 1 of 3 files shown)


aquadoggo/src/db/stores/document.rs
```diff
@@ -459,14 +464,59 @@ async fn insert_document_view(
...
-    .execute(tx)
-    .await
+    .execute(&mut *tx)
+    .await;
+
+    match result {
+        Ok(_) => Ok(true),
+        Err(err) => match err {
+            // Handle errors returned from the database
+            sqlx::Error::Database(_) => {
```
adzialocha, line 479: Can we make a query if that view exists before we insert, instead of handling an error?

I can see how the error approach makes one request less in case no data was in the database but it looks like it tries to insert on _all_ database errors?

I'm not against it, just want to compare the approaches.

adzialocha:
From our meeting today: We like handling the check on a higher level (inside the `reduce` task logic) than on the lower level store methods. It's:

1. Easier to read and follow what's going on (no hidden magic)
2. Stricter, might help us to find more bugs in other places with our logic

That's especially useful if we're working with the inner parts of `aquadoggo`.

sandreae:
This is ready, but the `ci` is failing cos of a crazy compiler error.... don't think it's related to this PR.

adzialocha (approved):
Great! I've removed the ON CONFLICT check in our SQL as well, since we handle the case in our code. Will merge now!

Issue #381, opened 2023-05-18
Cursor ignored during pagination


adzialocha:
Request 1:

```gql
{
  messages: all_chat_message_002055f4d34a8dd861270a0ebcb111acd1996ceda283dab1d0c09e59d5414e43a83f(
    orderBy: "timestamp"
    orderDirection: ASC
  ) {
    hasNextPage
    endCursor
    documents {
      cursor
      fields {
        text
        timestamp
      }
    }
  }
}
```

Result 1: 

```json
{
  "data": {
    "messages": {
      "hasNextPage": false,
      "endCursor": "33W9ydxf9SqZu3exkdHY9KZTkzYTk6WtCSHVbQxhcBYSS6GaBGgBQRdKfVzaKnsxUudM6R2yHCiarVGn9a3cGEsA",
      "documents": [
        {
          "cursor": "21Ca22qpjQkowN8X6sGusJKtMR8p8sQvYh7VMpFmwgdwf5ePVxzGPUVTc82k81dfcU2doWoDRAbSEpxDQGxuH89B",
          "fields": {
            "text": "1",
            "timestamp": 1684403519921
          }
        },
        {
          "cursor": "33W9ydxf9SqZu3exkdHY9KZTkzYTk6WtCSHVbQxhcBYSS6GaBGgBQRdKfVzaKnsxUudM6R2yHCiarVGn9a3cGEsA",
          "fields": {
            "text": "2",
            "timestamp": 1684403544448
          }
        }
      ]
    }
  }
}
```

Request 2 (using `cursor` of the last document in `after`):

```gql
{
  messages: all_chat_message_002055f4d34a8dd861270a0ebcb111acd1996ceda283dab1d0c09e59d5414e43a83f(
    after: "33W9ydxf9SqZu3exkdHY9KZTkzYTk6WtCSHVbQxhcBYSS6GaBGgBQRdKfVzaKnsxUudM6R2yHCiarVGn9a3cGEsA"
    orderBy: "timestamp"
    orderDirection: ASC
  ) {
    hasNextPage
    endCursor
    documents {
      cursor
      fields {
        text
        timestamp
      }
    }
  }
}
```

Result 2: Same as above, but should have been empty page.

adzialocha:
Suspicious behaviour, maybe related: The cursor _changes_ when `first` parameter gets changed to a number lower than the number of documents in that collection:

```gql
{
[...]

PR #412, opened 2023-06-20
Fix cursor pagination over ordered queries

adzialocha:
We paginate with the help of a unique `cursor`, there is one per document field.

```
Document {
  username <- Cursor #31c
  timestamp <- Cursor #21a
  message <- Cursor #d32
}

Document {
  username <- Cursor #a51
  ... and so on
}
```

When the client choses to order the resulting documents by some field (for example "timestamp") it is important that the comparison value for pagination comes from this field.

```
Query: Order "Chat" documents by "timestamp"

Document {
  username
  timestamp <- Compare
  message
}

Document {
  username
  timestamp <- Compare
  message
}

Document {
  username
  timestamp <- Compare
  message
}
```

At the same time we _need_ to always return the last `cursor` for each document to the client, since this is the point where we can safely paginate to the next document.

```
Document {
  username
  timestamp
  message <- Cursor
}

Document {
  username
  timestamp
  message <- Cursor
}

Document {
  username
  timestamp
  message <- Cursor
}
```

The bug resulted from us using the cursor to determine the field, which was only accidentally sometimes the same as chosen by the ordering.

Also, this fix uncovered a second bug which was that not always we returned the last cursor per document since the SQL sometimes orders them slightly different, indeterministicly.

This PR fixes the problem by:

1. Making sure that always the last field is selected as a cursor in `convert_rows` (before it's been a bit random, as the SQL query returns fields in different order sometimes)
2. Extending the pagination query to select the correct operation field based on the cursor and ordering setting

Closes https://github.com/p2panda/aquadoggo/issues/381

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/db/stores/query.rs
+++ b/aquadoggo/src/db/stores/query.rs
@@ -27,7 +27,7 @@ use crate::db::SqlStore;
 /// Configure query to select documents based on a relation list field.
 pub struct RelationList {
     /// View id of document which holds relation list field.
-    pub root: DocumentViewId,
+    pub root_view_id: DocumentViewId,
 
     /// Field which contains the relation list values.
     pub field: FieldName,
@@ -42,17 +42,17 @@ pub enum RelationListType {
 }
 
 impl RelationList {
-    pub fn new_unpinned(root: &DocumentViewId, field: &str) -> Self {
+    pub fn new_unpinned(root_view_id: &DocumentViewId, field: &str) -> Self {
         Self {
-            root: root.to_owned(),
+            root_view_id: root_view_id.to_owned(),
             field: field.to_string(),
             list_type: RelationListType::Unpinned,
         }
     }
 
-    pub fn new_pinned(root: &DocumentViewId, field: &str) -> Self {
+    pub fn new_pinned(root_view_id: &DocumentViewId, field: &str) -> Self {
         Self {
-            root: root.to_owned(),
+            root_view_id: root_view_id.to_owned(),
             field: field.to_string(),
             list_type: RelationListType::Pinned,
         }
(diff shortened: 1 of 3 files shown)


adzialocha:
I'm not really happy with this large query in the pagination logic. This abomination is _only_ for the pagination logic:

```sql
AND EXISTS (
    SELECT
        operation_fields_v1.value
    FROM
        operation_fields_v1
    WHERE
        operation_fields_v1.name = 'message'
        AND
            operation_fields_v1.operation_id = document_view_fields.operation_id
        AND
            (
                LOWER(operation_fields_v1.value) > (
                  SELECT
                      LOWER(operation_fields_v1.value)
                  FROM
                      operation_fields_v1
                      LEFT JOIN
                          document_view_fields
                          ON document_view_fields.operation_id = operation_fields_v1.operation_id
                  WHERE
                      document_view_fields.document_view_id = (
                          SELECT
                              document_view_fields.document_view_id
                          FROM
                              operation_fields_v1
                              LEFT JOIN
                                  document_view_fields
                                  ON document_view_fields.operation_id = operation_fields_v1.operation_id
                          WHERE
                              operation_fields_v1.cursor = 'fef4b411cd17ce1018d9720dc8d6135c5fd2f65e9e2f74fed5469b28051eab45'
                          LIMIT 1
                      )
                      AND operation_fields_v1.name = 'timestamp'
                  LIMIT 1
                  )
                OR
                (
                    LOWER(operation_fields_v1.value) = (
SELECT
    LOWER(operation_fields_v1.value)
FROM
    operation_fields_v1
    LEFT JOIN
        document_view_fields
        ON document_view_fields.operation_id = operation_fields_v1.operation_id
          WHERE
              document_view_fields.document_view_id = (
                  SELECT
                      document_view_fields.document_view_id
                  FROM
                      operation_fields_v1
                      LEFT JOIN
                          document_view_fields
                          ON document_view_fields.operation_id = operation_fields_v1.operation_id
                  WHERE
                      operation_fields_v1.cursor = 'fef4b411cd17ce1018d9720dc8d6135c5fd2f65e9e2f74fed5469b28051eab45'
                  LIMIT 1
              )
              AND operation_fields_v1.name = 'timestamp'
          LIMIT 1
          )
                    AND
                        operation_fields_v1.cursor > 'fef4b411cd17ce1018d9720dc8d6135c5fd2f65e9e2f74fed5469b28051eab45'
                )
            )
)
```

🤯 

These sub-queries return _always_ the same result. This is thanks to the `cursor` logic giving us a very precise way to point at data.

We can simplify this immensively by doing one query before the main query to retrieve the value of the cursors. I think this is even better for performance, otherwise we're running these expensive sub-queries for nothing.

The bug is fixed, but I'll refactor this code in the PR as well to make that extra query and safe brain cells and then it should be ready.

adzialocha:
I've added more docstrings to explain this whole `cursor` logic better and moved all sub-SELECT's related to pagination into separate SQL "pre" queries. Now the "main" query for pagination over ordered application fields looks like that:

```sql
AND EXISTS (
  SELECT
    operation_fields_v1.value
  FROM
    operation_fields_v1
  WHERE
    operation_fields_v1.name = 'name'
    AND operation_fields_v1.operation_id = document_view_fields.operation_id
    AND (
      LOWER(operation_fields_v1.value) > LOWER($1)
      OR (
        LOWER(operation_fields_v1.value) = LOWER($1)
        AND operation_fields_v1_list.cursor > 'c7679c82e8db45bf294c5a239484b108d98a2d3943f172582a7389e09b2327c4'
    )
)
```

adzialocha:
The PR is ready for review and merging now! I think I'll make a separate PR for moving the query.rs file into separate modules, it became VERY large (ticket here: https://github.com/p2panda/aquadoggo/issues/416)

aquadoggo/src/db/stores/query.rs
```diff
@@ -1494,6 +1754,121 @@ mod tests {
...
+        Order::new(&"timestamp".into(), &Direction::Ascending),
+        "message".into(),
+        vec![
+            "Hello, Panda!".into(),
+            "Oh, howdy, Pengi!".into(),
+            "How are you?".into(),
+            "I miss Pengolina. How about you?".into(),
+            "I am cute and very hungry".into(),
+            "(°◇°) !!".into(),
+        ],
```
sandreae, line 1768: :heart:

sandreae (approved):
Real tricky area, good work unwrapping it and thanks for presenting the solution in _as easy to understand as possible_ a way :+1:

PR #420, opened 2023-06-21
Fix race condition when check for existing view ids was too early

adzialocha:
This fixes a race condition where the check for duplicate document view ids in the database was too early.

The `reduce` task started, checked if a view id existed, if not, it happily went on, while another task could have created the view id in the meantime before we come to the point where we created the view id (again). Probably because of the ON CONFLICT clause we didn't notice that issue before.

## 📋 Checklist


sandreae (approved):
Nice, thanks!

Issue #401, opened 2023-06-01
Validation of entries and operations arriving via replication


sandreae:
Our handling of entries and operations via the `publish` api assumes that for any valid operation which arrives at the node we will be able to determine the `DocumentId` of for the document it's part of (by using the `previous` field to get the document from the store). The storage and validation API is built around this assumption.

When handling operations arriving via replication, then this may not be true though, as the entries + operations in multi-writer documents are unlikely to arrive in a way where the operation pointed to in `previous` is already in the store and attached to the corresponding document (the document may not exist at all yet).

There are quite a few ways to approach solving this problem at an implementation level (also at a spec level, but we'll mostly save that conversation for later).

*Edit: removed my initial ideas, need to think again...*

sandreae:
I've done some experiments with maintaining indexes for an operations topo sorted position in a document, this turned out to be quite simple. If we make use of this we could take the following approach:

- send operations + entries (batched per document) in their topo sorted order
- validate each operation against it's entry
- insert all entries into the db
- insert all operations in order into the db (this means we can expect the same guarantees around documents existing as we do in the `publish` api
- if any operation fails validation at this point, delete it's already inserted entry to clean up

We could achieve this with minimal changes to our existing storage API (I already did some small experiments here too). And it would mean we don't need any extra patterns for caching/processing not-yet-validated operations.

sandreae:
This is pretty elegant actually, makes use of our existing validation logic and we wouldn't need to change anything about the materialization service either. Shifting into `live` mode fits with this approach as well.

PR #432, opened 2023-06-29
Implement `Document` replication strategy

sandreae:
See here: https://github.com/p2panda/aquadoggo/issues/401 for background. This PR takes a little leap from there and directly implements a new strategy which replicates entries in their topologically sorted order, so we can validate and store entries using the existing `publish` api with only one minor change.

Still not quite ready for review, some open todos, but the general shape is good I think.

## Maintain indexes for all operations sorted position in their document 
We need this so that we can easily retrieve operations + entries in their topologically sorted order later on. This can only be done once an operation has been through the materializer, so `reduce` takes care of this for us. When an operation arrives on a node, before it has been materialized onto a document, it's `sorted_index` is `null`. 

## `get_entries_from_operation_index` on `SqlStore`

New method for retrieving all entries of a document from a specified operation index ("document height").

## Changes to `publish`

As we don't always replicate all logs between peers, and maybe not in the right order, then we need to already relax the requirement in `publish` for logs to be published into the "next" available space. This is implemented here: https://github.com/p2panda/p2panda/pull/498

## Introduce new replication mode and strategy
It made more sense to write a new strategy than to adapt our existing `LogHeight` strategy, so I implemented a the new `Document` strategy which has a new `Have` message. This was a good learning exercise, and it was really fun to see how easy it was to introduce a new strategy to our replication logic. It is a tangent from the primary concern right now which is sending entries and operation in topo order during replication, so if we feel it gets in the way/throws up too many new questions, we can factor it out and merge the new entry sending/consuming code into the existing `LogHeight` strategy. That feels a little hacky though (mixing document/log concerns quite a bit), but not the end of the world. 

### `HaveDocuments(Vec<(DocumentId, DocumentViewId)>)` 
This is the new `Have` message which `Document` strategy uses. In our original sketches for "replication by document view id" we believed we didn't need to send the `DocumentId` as well, but as it turns out, we do. The reason is:

PEER A document views: `[A1, B9, C3, D7]`
PEER B document views: `[A5, B2, C3, E4]`

If we compare the above sets from the perspective of `PEER B`, they can't identify the `DocumentId` for view ids `B9` or `D7`, as they have progressed further than our documents. This is a problem when we look at our own document `E4` as we don't know if `PEER A` knows about it, the unidentified documents (`B9` & `D7`) could be future versions of `E4`. Should we send them entries or not? We don't know.... The answer is to include the `DocumentId` as well. 

This makes the the `HaveDocument` message twice as large, which isn't nice. Interestingly, whether this is more or less data efficient probably depends mostly on the nature of the data we're handling. If it's many many documents by only a few authors, then `LogHeight` have messages will be much smaller, but if there are fewer documents and/or many many multi-writer authors, then `Document` have messages will be smaller.

### Diff document heights and collect entries for sending

New diffing logic for `Document` strategy, calculates "document height" for all local and remote documents, then compares them and calculates which entries we need to send.

## Changes to `send-to-node`

Funnily, an issue we already discussed last week came up here straight away when I was testing replication. As we were using the same key pair for publishing concurrent updates into different branches, then we break the expectation that entries will arrive in the correct sequence order. I fixed it in the demo by adding another key pair. We'd need to add this requirement to our spec too (https://github.com/p2panda/handbook/issues/284). 

## Tests

As I kept most of the `LogHeight` strategy intact (except `Ingest`) then I could add test cases for running many of the existing tests against both `LogHeight` and `Document` mode. Which is cool for catching any unexpected changes in behaviour.

## TODO
  
There are a few todos still around optimising the diff logic and handling one error. Also some teeeeeeeests.
## 📋 Checklist


sandreae:
(tests failing cos of a regression in the `PublicKey` serialization behaviour, being fixed elsewhere)

aquadoggo/src/db/models/operation.rs
```diff
@@ -25,6 +25,17 @@ pub struct OperationRow {
...
+    /// Index for the position of this operation once topological sorting of the operation graph
+    /// has been performed.
+    ///
+    /// This is an option as when an operation which is not appended to the tip of a
+    /// document graph is handled then a `reduce` materialization task will be issued and all
+    /// indexes for operations in the effected document re-calculated.
+    ///
+    /// If this value is None we can assume the operation has not been processed yet and we are
+    /// waiting for the reduce task to complete.
+    pub sorted_index: Option<i32>,
```
adzialocha, line 38: If this is `None` and we restart our node, would we still be able to materialize? I assume not, as the order (defined by the remote node we've received the data from) is lost?

adzialocha: If that is the case we maybe directly set the ordering on arrival but introduce a flag for if it has been materialized

adzialocha: Or we store the arrival order/ timestamp in our tasks table and reissue the tasks in the right order when the node restarts, that's probably the easiest

sandreae: > If this is None and we restart our node, would we still be able to materialize? I assume not, as the arrival order (defined by the remote node we've received the data from) is lost?

Actually, that's not the case. We don't use the order that the operations are supplied in for anything, we simply call `publish` in order and then as the subsequent `reduce` tasks complete, then `sorted_index` gets set for the new operations.

So we are actually leaning on our nice task re-run logic which already exists 👍

adzialocha: > Actually, that's not the case. We don't use the order that the operations are supplied in for anything except to call publish in the correct order. As the subsequent reduce tasks complete, sorted_index gets set for the new operations.

I mean the case when `reduce` has not been completed ..

Not sure if `tasks` currently returns tasks in the order they arrived on restart, that needs checking maybe.

adzialocha: Ah! Sorry, I see now. Somehow thought that publish comes later during replication 🙈

sandreae: > I mean the case when reduce has not been completed.

Hmm, yeh, I was meaning that too. Maybe I'm misunderstanding something. I thought:

- operations arrive, they're stored, `reduce` tasks issued (possibly de-duplication happens)
=====CRASH=====
- incomplete `reduce` tasks re-issued
- the order they're processed in doesn't matter, in any case all available operations will be included in the materialised document, and operations will have their `sorted_index` updated

sandreae:
I think this is good for review now @adzialocha !

sandreae:
A final comment from me: This PR is wildly more radical than it probably should be.... The issue it dealt with is that we had no good way to validate entries arriving via replication on nodes when using our original "log height" replication approach. It was rather cumbersome to implement sending operations in their sorted order on top of "log height" replication so I went ahead and implemented our "future" replication strategy which uses document identifiers and document heights to determine which entries+operations to replicate between nodes. This _was_ quite intuitive to implement, and ended with quite simple logic which fits our system (as we expected), but it really sends us off away from relying on log based replication, which is quite a sudden move, even though we expected it at some point. It also moves us closer to needing a set-reconciliation based strategy, as the `Have` messages will be bigger than with log based replication.

Ok, those are my final thoughts ;-p happy reviewing.

sandreae:
> I think this is good for review now @adzialocha !

I'm adding a few more tests for the diff logic actually... but still, that's all.

aquadoggo/src/replication/strategies/document/diff.rs
```diff
@@ -0,0 +1,307 @@
...
+                "Remote document height: {} <DocumentViewId {}> {remote_height}",
+                local_document.id().display(),
+                remote_document_view_id.display()
+            );
+
+            // Calculate the local document height.
+            //
+            // @TODO: This could be made more efficient if we store the current document height in the
+            // store.
+            let (local_height, _) =
```
sandreae, line 161: Ah yeh, missing some logic around here which accounts for document view ids which contain multiple operation ids.

sandreae: When the local node has multiple tips, and the remote doesn't, it can be that the document heights appear to be the same but they're actually not.

sandreae:
UPDATE:

There was flaw in diff logic of this document replication strategy which meant comparing documents, where one was missing a branch, wouldn't notice this difference. This is obviously quite a serious flaw :sweat_smile:. I made a few changes to account for these situations as I still think it's worth seeing how far we can go with this strategy, and there was definitely some interesting learning. With the changes outlined bellow it now works (can diff documents across nodes and calculate required entries to send). I don't think it's obviously "better" than our existing log-height replication approach though (although it's interesting in that it doesn't require logs at all) and so I think this PR has been good research and is of interest, likely something we don't want/need to merge to main. 

The changes required for the diffing logic to work were:
- maintain `DocumentViewId` for every operation in the store, which points to the view of the document when that operation was materialized. This allows us to know which unmerged branches were known to the node at that particular step (operation) in it's materialization process.
- during diffing a remotes document with our own local one (identified by their `DocumentId` and `DocumentViewId`) we can now check that all branches were known about before sending our respone.
- Were all branches known about?:
  - Yes: send only the operations required to update the document to current state
  - No: send all operations for the document as we need to make sure all missed branches are accounted for

aquadoggo/src/replication/strategies/document/diff.rs
sandreae: See comment below: https://github.com/p2panda/aquadoggo/pull/432#issuecomment-1627651760

Issue #418, opened 2023-06-21
`reduce` tasks gets into long materialization loop


adzialocha:
Since we changed https://github.com/p2panda/aquadoggo/pull/413 I observe this behaviour of a very long loop of trying to re-schedule building that view id:

```
...
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::tasks::reduce] Find document for view with id: 002023bb78c051a8d73aa26094b60d05ef3ef8fba8008f62d2c30f1fe8d9dcf87925
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::tasks::reduce] Document materialized to view with id: 002023bb78c051a8d73aa26094b60d05ef3ef8fba8008f62d2c30f1fe8d9dcf87925
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::tasks::reduce] Document <DocumentId f87925> for view not materialized yet, reissuing current reduce task for f87925
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::worker] Sending materializer reduce task with input <TaskInput -/002023bb78c051a8d73aa26094b60d05ef3ef8fba8008f62d2c30f1fe8d9dcf87925> to the task queue.
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::tasks::reduce] Working on <TaskInput -/002023bb78c051a8d73aa26094b60d05ef3ef8fba8008f62d2c30f1fe8d9dcf87925>
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::tasks::reduce] Find document for view with id: 002023bb78c051a8d73aa26094b60d05ef3ef8fba8008f62d2c30f1fe8d9dcf87925
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::tasks::reduce] Document materialized to view with id: 002023bb78c051a8d73aa26094b60d05ef3ef8fba8008f62d2c30f1fe8d9dcf87925
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::tasks::reduce] Document <DocumentId f87925> for view not materialized yet, reissuing current reduce task for f87925
[2023-06-21T14:19:29Z DEBUG aquadoggo::materializer::worker] Sending materializer reduce task with input <TaskInput -/002023bb78c051a8d73aa26094b60d05ef3ef8fba8008f62d2c30f1fe8d9dcf87925> to the task queue.
[...]

Issue #411, opened 2023-06-20
Materialize missing document view ids when operations arrive late


adzialocha:
The following scenario is not handled by our materializer right now:

We have a Document `A` with the latest view `A1` and Document `B` with the latest view `B3`.

1. Document `A` has a pinned relation to document view `B2`
2. Document `A` gets materialized but the view id `B2` does not exist yet, our node doesn't have any operations of `B` yet
3. _All_ operations ever created of document `B` arrive, we materialize it and create document `B` and document view `B3`, since this is the latest version of it
4. We query document `A` and the pinned relation field still returns nothing

This shouldn't be a critical error as nothing crashes in this scenario but it is still wrong.

To solve this we might need to look at reverse relations in the `dependency` task, that is, checking which child documents have a pinned relation using the schema id of the parent document.

Issue #152, opened 2022-06-13
Dependency task should also look at `SchemaId` relations of the document view


adzialocha:
Currently the dependency task implementation only looks at pinned / unpinned relations of the current document view, but we should also take the schema id into account (basically making the application document view a "parent" of the schema definition document).

PR #434, opened 2023-07-03
Reverse lookup for pinned relations in dependency task

adzialocha:
See diagram for more details: https://whimsical.com/materializer-DbV6FUs51pqhouKff7Q546

Closes: https://github.com/p2panda/aquadoggo/issues/418 and https://github.com/p2panda/aquadoggo/issues/411 and https://github.com/p2panda/aquadoggo/issues/152

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/materializer/tasks/dependency.rs
+++ b/aquadoggo/src/materializer/tasks/dependency.rs
@@ -110,22 +110,18 @@ pub async fn dependency_task(context: Context, input: TaskInput) -> TaskResult<T
     }
 
     // Construct additional tasks if the task input matches certain system schemas and all
-    // dependencies have been reduced.
+    // dependencies have been reduced
     let all_dependencies_met = !next_tasks.iter().any(|task| task.is_some());
     if all_dependencies_met {
-        // Helper that returns a schema task for the current task input.
-        let schema_task = || {
-            Some(Task::new(
-                "schema",
-                TaskInput::DocumentViewId(document_view.id().clone()),
-            ))
-        };
-
         match document.schema_id() {
             // Start `schema` task when a schema (field) definition view is completed with
             // dependencies
-            SchemaId::SchemaDefinition(_) => next_tasks.push(schema_task()),
-            SchemaId::SchemaFieldDefinition(_) => next_tasks.push(schema_task()),
+            SchemaId::SchemaDefinition(_) | SchemaId::SchemaFieldDefinition(_) => {
+                next_tasks.push(Some(Task::new(
+                    "schema",
+                    TaskInput::DocumentViewId(document_view.id().clone()),
+                )));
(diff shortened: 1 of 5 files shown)


aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -100,42 +104,44 @@ pub async fn dependency_task(context: Context, input: TaskInput) -> TaskResult<T
...
+    // Now we check all the "parent" or "inverse" relations, that is _other_ documents pointing at
+    // the one we're currently looking at
+    let mut reverse_tasks = get_inverse_relation_tasks(&context, document.schema_id()).await?;
+    next_tasks.append(&mut reverse_tasks);
+
     // Construct additional tasks if the task input matches certain system schemas and all
-    // dependencies have been reduced.
-    let all_dependencies_met = !next_tasks.iter().any(|task| task.is_some());
+    // dependencies have been reduced
+    let all_dependencies_met = next_tasks.is_empty();
```
sandreae, line 127: Now that any "reverse_tasks" are included in the `next_tasks` is this check performing as we'd expect? I don't see a reason not to issue a `schema` task if there are reverse tasks still to work on. As I understand it, we only want to make sure all child/direct relation dependencies are met.

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -166,18 +172,79 @@ async fn get_relation_task(
     }
 }
 
+/// Returns _reduce_ tasks for every document which has a pinned relation or pinned relation list
```
sandreae, line 175: I think this should say "dependency" tasks.

aquadoggo/src/materializer/tasks/dependency.rs
```diff
@@ -654,4 +722,376 @@ mod tests {
             assert_eq!(schema_tasks, expected_schema_tasks);
         });
     }
+
+    #[rstest]
+    fn materialise_late_views(key_pair: KeyPair) {
```
sandreae, line 727: Haha, epic test :rofl: love it.

sandreae (changes requested):
All looking good except for one point I commented on.

aquadoggo/src/materializer/tasks/dependency.rs
adzialocha: Right, when we're looking at a `schema_definition` document that's right, I'll fix this!

aquadoggo/src/materializer/tasks/dependency.rs
adzialocha: Hehe, it has a lot of boilerplate because I needed to enforce the order of things being processed without also adding the whole replication system 🙈

PR #438, opened 2023-07-07
Maintain sorted operation indexes

sandreae:
We want to persist and maintain sorted position indexes for all operations which have been materialized into a document. This index is the operations position once topological sorting has occurred and can be understood as the `sorted_index` of a particular operation, or also the documents "depth" at that point in it's history.

This is done by introducing the following:
- add a `sorted_index` column to the `operations_v1` table
- update all operation models and data types accordingly
- introduce a `update_operation_index` method to the `OperationStore`
- in a `reduce` task update the `sorted_index` of all operations in a document

Note, there is a period of time when an operation has been inserted into the store, but it's `sorted_index` is `None` because materialization has not been performed yet. As we have a "retry" system already in-place to account for a node shutting down unexpectedly before tasks could be completed, then we can be confident all operations will be processed and have their `sorted_index` set eventually.

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Naive protocol replication [#380](https://github.com/p2panda/aquadoggo/pull/380)
 - Integrate replication manager with networking stack [#387](https://github.com/p2panda/aquadoggo/pull/387) 🥞
 - Reverse lookup for pinned relations in dependency task  [#434](https://github.com/p2panda/aquadoggo/pull/434)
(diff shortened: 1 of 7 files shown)


adzialocha:
Great!!

> Note, there is a period of time when an operation has been inserted into the store, but it's sorted_index is None because materialization has not been performed yet. As we have a "retry" system already in-place to account for a node shutting down unexpectedly before tasks could be completed, then we can be confident all operations will be processed and have their sorted_index set eventually.

That's half working actually, but with this PR we have everything in place I think to help us with it: https://github.com/p2panda/aquadoggo/issues/441

PR #444, opened 2023-07-13
Implement `dialer` behaviour

sandreae:
Introduce a `Dialer` behaviour which:

- handles new dial attempts for known peers or peers which disconnect unexpectedly
- when a dial attempt fails retries according to a configured backoff schedule and retry limit

## 📋 Checklist

- [x] ~Add tests that cover your changes~ I'm struggling to find a meaningful way to test this, open to any ideas, otherwise it can go in without tests..... (I tested locally and expected behaviour occurs)

### DIFF
--- a/aquadoggo/src/network/dialer/behaviour.rs
+++ b/aquadoggo/src/network/dialer/behaviour.rs
@@ -17,7 +17,7 @@ use std::time::{Duration, Instant};
 
 const RETRY_LIMIT: u32 = 8;
 const BACKOFF_SEC_MIN: u64 = 1;
-const BACKOFF_SEC_MAX: u64 = 10;
+const BACKOFF_SEC_MAX: u64 = 60;
 
 /// Events which are sent to the swarm from the dialer.
 #[derive(Clone, Debug)]
--- a/aquadoggo/src/network/dialer/behaviour.rs
+++ b/aquadoggo/src/network/dialer/behaviour.rs
@@ -5,30 +5,45 @@ use std::task::{Context, Poll};
 
 use libp2p::core::Endpoint;
 use libp2p::swarm::derive_prelude::ConnectionEstablished;
-use libp2p::swarm::dial_opts::{DialOpts, PeerCondition};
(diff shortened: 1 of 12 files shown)


aquadoggo/src/network/dialer/behaviour.rs
```diff
@@ -0,0 +1,246 @@
...
+        Ok(DummyConnectionHandler)
+    }
+
+    fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>) {
+        match event {
+            FromSwarm::ConnectionEstablished(ConnectionEstablished { peer_id, .. }) => {
+                self.on_connection_established(peer_id);
+            }
+            FromSwarm::ConnectionClosed(ConnectionClosed { peer_id, .. }) => {
+                self.on_connection_closed(&peer_id);
```
adzialocha, line 182: This would try to redial even if the other peer closed the connection for good, right? Can we make this happen only on an error? Think for this we might need to call a behavior method from the swarm.

adzialocha: Or is this dial condition taking care of that?

sandreae: Good point! I've accounted for this now with the re-write.

aquadoggo/src/network/dialer/behaviour.rs
```diff
@@ -0,0 +1,248 @@
...
+use libp2p::core::Endpoint;
+use libp2p::swarm::derive_prelude::ConnectionEstablished;
+use libp2p::swarm::dummy::ConnectionHandler as DummyConnectionHandler;
+use libp2p::swarm::{
+    ConnectionDenied, ConnectionId, DialFailure, FromSwarm, NetworkBehaviour, PollParameters,
+    THandler, THandlerInEvent, THandlerOutEvent, ToSwarm,
+};
+use libp2p::{Multiaddr, PeerId};
+use log::{debug, warn};
+use std::time::{Duration, Instant};
```
adzialocha: Import grouping

adzialocha (approved):

PR #467, opened 2023-07-26
Reduce chatty warnings in network and replication services

sandreae:
- [x] don't `warn!` (just `debug!`) when a session wasn't found for a replication message, this has normally been followed by the actual error which closed the session
- [x] reduce most logging which we are bubbling up from `libp2p` in the network service to `debug!` (only `warn!` when it's a connection closure event that we are explicitly responding to in panda land) 
- [x] don't `warn` when dial attempts fail, this is normally because there is an existing dial attempt in progress, it's not an actual connection error, these get picked up elsewhere.

## 📋 Checklist

- [x] ~Add tests that cover your changes~


sandreae:
Some of these logs were. Really. Annoying.

sandreae:
We can also do a lot better than this, but it's an improvement at least.

PR #485, opened 2023-08-04
More finely grained `TaskInput` variants

sandreae:
In "blob" tasks we need to be able to determine if the task we're working on is relating to the view of a pinned document view, or the current view of a document (the current input for both of these is `DocumentViewId`).

This is because we want to forward HTTP requests from `blobs/document_id` to `blobs/document_id/document_view_id`, and a good way of doing this would be through creating symlinks on the filesystem. If we know the "blob" task is dealing with a current view then we can easily do it at this point.

The idea in this PR is to make the `TaskInput` enum variants more finely grained, so we can determine between these two types of task input.

## 📋 Checklist


adzialocha:
> This is because we want to forward HTTP requests from blobs/document_id to blobs/document_id/document_view_id, and a good way of doing this would be through creating symlinks on the filesystem. If we know the "blob" task is dealing with a current view then we can easily do it at this point.

With the danger of not understanding the problem fully (travelling brain), this smells slightly like trying to combine two things which do not need to know about each other: Why does the "blob" task need to know about symlinks for the HTTP server?

Could your problem rather be solved with resolving symlinks when doing the HTTP query? Surely that might mean to not use the convenient "static file hosting" feature, but maybe its required to build our own redirection logic? Maybe in form of a middleware?

Intuitively I'd look for solving that problem there and maybe adding this slightly "alien" input type to the materializer is a sign that this is the "wrong" place?

That's my first idea, if you say there's no other way, I think a bit more about the naming.

PR #493, opened 2023-08-08
"blob" materializer task

sandreae:
Introduces a new "blob" task to the materializer service which:
- is issued once a dependency task finishes on a `blob_v1` or `blob_piece_v1` document
- if the document in question is part of a completed blob (all dependencies met, this isn't always true when a `blob_piece_v1` document triggered the "blob" task) then we persist the blob view to the filesystem at the expected path `../<blob_document_id>/<blob_view_id>`
- we also create a symlink from the expected path to the blob when served by it's document id ../documents/<blob_document_id>`.
  
## Next steps

- would be good to have a `blob` module which contains all `io` stuff 

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/config.rs
+++ b/aquadoggo/src/config.rs
@@ -19,6 +19,8 @@ const DEFAULT_SQLITE_NAME: &str = "aquadoggo-node.sqlite3";
 /// Blobs directory
 pub const BLOBS_DIR_NAME: &str = "blobs";
 
+pub const BLOBS_SYMLINK_DIR_NAME: &str = "documents";
+
 /// Configuration object holding all important variables throughout the application.
 ///
 /// Each configuration also assures that a data directory exists on the host machine where database
@@ -81,7 +83,7 @@ impl Configuration {
         });
 
         // Create folders when they don't exist yet
-        fs::create_dir_all(base_path.join(BLOBS_DIR_NAME))?;
+        fs::create_dir_all(base_path.join(BLOBS_DIR_NAME).join(BLOBS_SYMLINK_DIR_NAME))?;
 
         Ok(base_path)
     }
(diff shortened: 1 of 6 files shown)


aquadoggo/src/materializer/tasks/blob.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    Ok(None)
+}
+
+/// Retrieve blobs that use the targeted blob piece as one of their fields.
+async fn get_related_blobs(
+    target_blob_piece: &DocumentViewId,
+    context: &Context,
+) -> Result<Vec<StorageDocument>, TaskError> {
+    // Retrieve all blob documents from the store
```
sandreae, line 123: Use custom SQL query here.

aquadoggo/src/config.rs
```diff
@@ -19,6 +19,8 @@ const DEFAULT_SQLITE_NAME: &str = "aquadoggo-node.sqlite3";
 /// Blobs directory
 pub const BLOBS_DIR_NAME: &str = "blobs";
 
+pub const BLOBS_SYMLINK_DIR_NAME: &str = "documents";
```
sandreae, line 22: Is this a good name? (and give it a comment)

adzialocha: Would be nice if we could get rid of this extra path and use our own router to serve static files.

I would think of something like this:

Route: `/blobs/:documentId/?:documentViewId`

With these two parameters `documentId` and optionally `documentViewId` we would resolve the right file on the filesystem and return the bytes to the HTTP client. Additionally we would also do a database lookup, finding out the MIME type and adding that to the HTTP response.

adzialocha (approved):
Great! Looking all good.

I would consider adding our own routes and handlers for the `/blobs` paths. Both for MIME types and helping with the symlinking workaround. I'm fine to handle this in a separate PR though.

aquadoggo/src/materializer/tasks/blob.rs
sandreae: There might be some re-use possibilities once both open blob PRs are merged so I will wait on refactoring this.

aquadoggo/src/config.rs
sandreae: Yep, will add a note and tackle this in a separate PR.

PR #496, opened 2023-08-09
"prune" materializer task

sandreae:
Introduces a "prune" materialization task which:
- is issued whenever a "reduce" task completes when the input was a `DocumentId` (this is a new current view)
- the "prune" task looks at the updated document and deletes any rows from the `document_views` and `document_view_fields` table where: they are not the current view AND they are not targeted from a pinned relation of any other document

I also had to make some changes to `prune_document_views` method which was introduced onto the store in a previous PR cos it was not doing what it's supposed to....

## 📋 Checklist


sandreae:
This work is being continued in a new branch here: https://github.com/p2panda/aquadoggo/pull/500

PR #499, opened 2023-08-11
Purge documents

sandreae:
## 📋 Checklist

- [ ] Add tests that cover your changes
- [ ] Add this PR to the _Unreleased_ section in `CHANGELOG.md`
- [ ] Link this PR to any issues it closes
- [ ] New files contain a SPDX license header

PR #500, opened 2023-08-12
Task for automatic garbage collection of unused documents and views

sandreae:
_i think this PR is made up of just 10% actual new code and 90% tests....._

> I reverted changes implemented in https://github.com/p2panda/aquadoggo/pull/499 and https://github.com/p2panda/aquadoggo/pull/496 in order to combine them into one PR (this one) as pruning and purging documents can be handled more elegantly in a single task, now called "garbage_collection".

Introduces a new "garbage_collection" task to the materializer service. It is concerned with two different connected but discreet functions:

1. Deleting document views: when a document view is no longer pinned by any document, and is no documents current view, it is deleted from the store.
2. Purging blobs: when a blob document is no longer related to by any other document (pinned or unpinned relation) then it's views, operations, and entries can are deleted from the store

The purging only occurs for blobs, but this functionality could be re-used in the future to handle other kind of "dependent" documents which should only be kept alive when related to by other documents. 

One point to note is that no rows are removed from the `logs` table as we still need to block off used log ids.

**Next steps:**

- Once this PR and the "blob" task PR are merged we need to hook up the garbage collection logic to also delete blobs which were materialized to the filesystem: https://github.com/p2panda/aquadoggo/issues/517

## 📋 Checklist


sandreae:
More PostgreSQL woes.... this time it doesn't like how we do triggers, it wants us to define a function, but SQLite doesn't do functions....

aquadoggo/src/db/stores/blob.rs
```diff
@@ -53,6 +55,98 @@ impl SqlStore {
...
+
+    /// Purge blob data from the node _if_ it is not related to from another document.
+    pub async fn purge_blob(&self, document_id: &DocumentId) -> Result<(), SqlStoreError> {
+        // Collect the view id of any existing document views which contain a relation to the blob
+        // which is the purge target.
+        let blob_reverse_relations = reverse_relations(&self.pool, document_id, None).await?;
+
+        // If there are no documents referring to the blob then we continue with the purge.
+        if blob_reverse_relations.is_empty() {
+            // Collect the document view ids of all pieces this blob has ever referred to in it's
```
sandreae, line 67: Add a test to check purging does remove all pieces (even if the blob was updated).

sandreae: We don't _need_ to support blob updating, but so far I think we do.

aquadoggo/src/db/stores/blob.rs
```diff
@@ -53,6 +55,98 @@ impl SqlStore {
...
+                AND
+                    operation_fields_v1.name = 'pieces'
+                ",
+            )
+            .bind(document_id.to_string())
+            .fetch_all(&self.pool)
+            .await
+            .map_err(|e| SqlStoreError::Transaction(e.to_string()))?;
+
+            self.purge_document(document_id).await?;
```
sandreae, line 90: More comments here.

aquadoggo/src/db/stores/document.rs
```diff
@@ -366,76 +381,161 @@ impl SqlStore {
...
+        )
+        .bind(document_view_id.to_string())
+        .fetch_optional(&self.pool)
+        .await
+        .map_err(|err| DocumentStorageError::FatalStorageError(err.to_string()))?;
+
+        Ok(document_view_id.is_some())
+    }
+
+    pub async fn purge_document(
```
sandreae, line 525: Doc string.

aquadoggo/src/materializer/tasks/garbage_collection.rs
```diff
@@ -0,0 +1,649 @@
...
+                if view_deleted {
+                    debug!("Deleted view: {}", document_view_id);
+                    deleted_views_count += 1;
+                    all_effected_child_relations.extend(effected_child_relations);
+                } else {
+                    debug!("Did not delete view: {}", document_view_id);
+                }
+            }
+
+            // If the number of deleted views equals the total existing views, then there is a
```
sandreae, line 80: Correct comment: add "minus one"

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Serve static files from `blobs` directory [#480](https://github.com/p2panda/aquadoggo/pull/480)
 - Add method to store for pruning document views [#491](https://github.com/p2panda/aquadoggo/pull/491)
 - Introduce `BlobStore` [#484](https://github.com/p2panda/aquadoggo/pull/484)
+- `garbage_collection` materialization task [#500](https://github.com/p2panda/aquadoggo/pull/500)
```
adzialocha, line 36: Think its nice to explain a bit more _what_ got added here

adzialocha: .. the name of this PR is already much better: "Task for automatic garbage collection of unused documents and views"

aquadoggo/migrations/20230811201924_purge-document-trigger.sql
```diff
@@ -0,0 +1,14 @@
...
+BEGIN
+    DELETE FROM entries 
+    WHERE entries.entry_hash IN (
+        SELECT operations_v1.operation_id FROM operations_v1
+        WHERE operations_v1.document_id = OLD.document_id
+    );
+    
+    DELETE FROM operations_v1
+    WHERE operations_v1.document_id = OLD.document_id;
+END;
```
adzialocha, line 14: This was the part which was not supported by PostgreSQL, as we said in our last meeting, lets move it into separate SQL queries in Rust to allow compatibility (for now ..).

aquadoggo/src/db/errors.rs
```diff
@@ -17,6 +17,10 @@ pub enum SqlStoreError {
     /// Error returned from BlobStore.
     #[error(transparent)]
     BlobStoreError(#[from] BlobStoreError),
+
+    /// Error returned from `DocumentStore` methods.
+    #[error(transparent)]
+    DocumentStorage(#[from] DocumentStorageError),
```
adzialocha, line 23: Ah funny, I realized that the errors always end on "Storage" why the rest of the APIs always talk about "Store". Not important, just wanted to mark that.

aquadoggo/src/db/stores/document.rs
```diff
@@ -366,76 +381,161 @@ impl SqlStore {
...
                 WHERE
                     document_views.document_view_id = $1
                 AND NOT EXISTS (
-                    SELECT * FROM operation_fields_v1
+                    SELECT 
+                        document_view_fields.document_view_id 
+                    FROM 
+                        document_view_fields
+                    LEFT JOIN
+                        {OPERATION_FIELDS}
```
adzialocha, line 479: I would write the SQL query as-is in here, its harder to read with these constant values, same in all the other spots.

adzialocha (changes requested):

aquadoggo/src/db/errors.rs
sandreae: Oh yeh :sweat_smile: we can make this more consistent in a follow-up PR if we like.

Issue #423, opened 2023-06-22
Add holepunching via libp2p's DCUTR over QUIC


adzialocha:
This is now implemented in `libp2p-v0.52.0`

See example here: https://github.com/libp2p/rust-libp2p/tree/libp2p-v0.52.0/examples/dcutr

Issue #361, opened 2023-05-08
libp2p: AutoNAT on QUIC falsely reports public NAT status


sandreae:
Might effect us as we do exactly this: https://github.com/libp2p/rust-libp2p/issues/3900

adzialocha:
Maybe that behaviour is too magic anyway and we should rather aim for a nice `config.toml`?

PR #502, opened 2023-08-17
Fix relayed connections, add DCUtR and reduce CLI args

sandreae:
I had a few goals with this PR:

1. Troubleshoot and get relayed connection establishment to be more reliable
2. Reduce the number of network related CLI args 
3. Integrate and test if DCUtR now works
4. Choose random QUIC and HTTP ports when they're taken

This all came with a general refactor, especially of the event loop, which is hopefully now a little easier to follow. The answer to 1. seems to have been to introduce a kind of handshake phase to the initiation process if a relay address has been given. This waits on identify messages to/from the relay, and successful listening on the circuit address before starting any other processes. Good news is, with this, relayed connections seem to be very quick and reliable :star:

I introduced the DCUtR behaviour but I haven't been able to observe if it works yet, this could just be my particular NAT setup not allowing holepunching though. There are upgrade requests passing back and forth, but the upgrade itself fails. This doesn't seem to have any adverse effect on the existing relayed connection though.

Lastly, I removed the `dial` behaviour which was responsible for redialing peers if they disconnect unexpectedly. This was partly as I thought it might be effecting the DCUtR processes, and partly because I wanted to in general reduce complicating the connection logic at all. It may come back later, lets see (also connections seem pretty reliable anyway).  

Closes: https://github.com/p2panda/aquadoggo/issues/423 and https://github.com/p2panda/aquadoggo/issues/361

## 📋 Checklist


aquadoggo/src/network/behaviour.rs
```diff
@@ -166,29 +130,104 @@ impl Behaviour {
...
             relay_server: relay_server.into(),
+            dcutr: dcutr.into(),
             peers,
-            dialer,
         })
     }
 }
+
+#[derive(Debug)]
+pub enum Event {
```
adzialocha, line 169: @sandreae, I remember you mentioned some weeks ago why this wrapper enum was required. Was it when we have two swarms? Is this still necessary?

aquadoggo/src/network/peers/behaviour.rs
```diff
@@ -110,32 +108,39 @@ impl Behaviour {
...
+
+    /// Stop the peers behaviour from running, it won't handle any connection events or received messages.
+    pub fn stop(&mut self) {
+        self.running = false
+    }
+
+    /// Restart the peers behaviour.
+    pub fn restart(&mut self) {
+        self.running = true
     }
```
adzialocha, line 132: Ah, you already have that thing implemented we talked about during our meeting today! 😆  (I thought that `stop` and `restart` are some native network behaviour methods). Are you concerned about the established connections or what was the problem again?

aquadoggo/src/network/service.rs
```diff
@@ -16,111 +16,265 @@ use tokio_stream::StreamExt;
...
 use crate::context::Context;
 use crate::manager::{ServiceReadySender, Shutdown};
-use crate::network::behaviour::{Behaviour, BehaviourEvent};
+use crate::network::behaviour::{Event, P2pandaBehaviour};
 use crate::network::config::NODE_NAMESPACE;
-use crate::network::{dialer, identity, peers, swarm, NetworkConfiguration, ShutdownHandler};
+use crate::network::{identity, peers, swarm, NetworkConfiguration, ShutdownHandler};
 
-/// Network service that configures and deploys a libp2p network swarm over QUIC transports.
+static DEFAULT_QUIC_PORT: u16 = 2022;
```
adzialocha, line 23: I would not have expected the default values to be defined in the service but somewhere close to `NetworkConfiguration`.

aquadoggo/src/network/peers/behaviour.rs
sandreae: The problem is that during the client-relay handshake phase (where identify info is exchanged and circuit relay established) the peers behaviour understands any connections as being new replication peers and so initiates sync sessions and sends begins sending messages. I wanted to avoid this getting mixed up in the initial node setup phase and so added these methods to basically pause replication until we want it to start (after initialisation).

adzialocha: Okay, cool. Why is then the second dial required?

sandreae: That triggers both peers to start replicating with each other. It's not beautiful, but it works!

aquadoggo/src/network/behaviour.rs
sandreae: This is a pattern which I copied cos I saw it in many other projects using libp2p, and also their own examples. To be honest, I didn't think deeply about it 😅

adzialocha: I've just played with it, don't think it's necessary, but it makes the event types look nicer!

aquadoggo/src/network/peers/behaviour.rs
adzialocha: Okay, I'm adding a TODO then :+1:

adzialocha:
In mDNS I observe that we have multiple connections now, I guess that's due to running TCP and QUIC at the same time, right?

sandreae:
> In mDNS I observe that we have multiple (up to 4) connections / replication sessions with the same peer now, I guess that's due to running TCP and QUIC at the same time, right?

Ah yeh, possibly. I can test the successful DCUtR without listening on TCP. If it still works we can remove listening on TCP for the "client" nodes.

mycognosist:
@sandreae 

Thanks for the review request! I'm planning to read through everything and provide feedback tomorrow.

sandreae:
> @sandreae
> 
> Thanks for the review request! I'm planning to read through everything and provide feedback tomorrow.

Great, thanks!

mycognosist (approved):
Sweet! All looks good to me. Well done!

Great to hear that the relayed connections are now reliable and speedy. The addition of the reservation timeouts is a nice touch and should provide significant stability for long-lived sessions / connections.

Nice to see the CLI args pared-down further; always love to see a codebase getting smaller :)

Give me a shout if you ever want more hands on deck to test DCUtR.

sandreae:
> Sweet! All looks good to me. Well done!
> 
> Great to hear that the relayed connections are now reliable and speedy. The addition of the reservation timeouts is a nice touch and should provide significant stability for long-lived sessions / connections.
> 
> Nice to see the CLI args pared-down further; always love to see a codebase getting smaller :)
> 
> Give me a shout if you ever want more hands on deck to test DCUtR.

Thanks @mycognosist, glad you're happy with the direction. It it started with your good work :smile: !

Yeh, always satisfying to reduce code, we started with "we have all the options, yey!" and ending up at "really we only need these options for now, yey!"

Would defo be good to do some more DCUtR at some point. Happily, when it doesn't succeed, there are no negative effects on the currently relayed connection, so really it's a bonus. But it would be nice to have a rough idea of when it is likely to succeed.

adzialocha:
> > In mDNS I observe that we have multiple (up to 4) connections / replication sessions with the same peer now, I guess that's due to running TCP and QUIC at the same time, right?
> 
> This was actually cos we were missing dial opts with conditions (`dialer` used to contain these) for dialing peers discovered over mDNS. We only want to dial when the peer is disconnected _and_ not already being dialed. Additionally we dial concurrency to `1`. With this i observe max one connection being established as a listener or dialer (we still can't avoid these two connections being established). 

Nice one! That looks much more transparent.

PR #514, opened 2023-08-18
`blob_v1` and `blob_piece_v1` dependency replication

sandreae:
`blob_v1` and `blob_piece_v1` documents should only be replicated if they are related to from another document which is included because of schema specified in the target set. For example:

- `TargetSet([img_0020, blob_v1])` would included all `img_0020` documents and any `blob_v1` documents they relate to
- `TargetSet([img_0020, blob_v1, blob_piece_v1])` would included all `img_0020` documents and any `blob_v1` documents they relate to, and any `blob_piece_v1` documents _they_ relate to
- `TargetSet([img_0020])` would included all `img_0020` documents and nothing else
- `TargetSet([blob_v1, blob_piece_v1])` would not include anything

Next step is to add a config flag which adds the correct blob system schema to the target set for you: https://github.com/p2panda/aquadoggo/issues/564 

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Task for automatic garbage collection of unused documents and views [#500](https://github.com/p2panda/aquadoggo/pull/500)
 - Blobs directory configuration [#549](https://github.com/p2panda/aquadoggo/pull/549)
 - Integrate `Bytes` operation value [554](https://github.com/p2panda/aquadoggo/pull/554/)
+- Implement dependency replication for `blob_v1` and `blob_piece_v1` documents [#514](https://github.com/p2panda/aquadoggo/pull/514)
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


adzialocha (approved):
Thank you! I find the new "rules" quite intuitive. Obviously that will change the "magic demo" character of `aquadoggo` slightly, as you can't just replicate blobs anymore like that. You'll need a schema first (aka `fishy` to create one) which has a relation to a blob!

sandreae:
> Thank you! I find the new "rules" quite intuitive. Obviously that will change the "magic demo" character of `aquadoggo` slightly, as you can't just replicate blobs anymore like that. You'll need a schema first (aka `fishy` to create one) which has a relation to a blob!

Thanks 🙏 glad it's intuitive. 

Yeh, I'd considered how this effects our "it just works" node setup. I thought about adding a rule that if you only have `blob_v1` and `blob_piece_v1` then you magically replicate blobs again. Wasn't sure if that actually just made things more complicated though 🤔 once we have multiple sync sessions between peers this can be nice addition. Probably with config changes we can have a nice "it just works" setup again.

sandreae:
We could easily have an `i_want_all_the_blobs` flag which over-rides this behaviour.

adzialocha:
> We could easily have an `i_want_all_the_blobs` flag which over-rides this behaviour.

Yeah, let's merge this for now, I think we need some more time to think about a good UI for configuration.

PR #515, opened 2023-08-20
Announce supported schema ids in network before replication

### LINKED ISSUE #405: Announcement messages

adzialocha:
This PR introduces our concept of "Announcements" (read more here: https://p2panda.org/specification/replication/#1-announcement).

There's a specification change involved: We need to give `Announce` messages their own message type. I've took `0` for them and changed the ones for `SyncRequest`, `SyncDone` and `Entry` as well. Related PR: https://github.com/p2panda/handbook/pull/290

Closes #405 

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/network/peers/behaviour.rs
+++ b/aquadoggo/src/network/peers/behaviour.rs
@@ -12,13 +12,12 @@ use libp2p::swarm::{
 use libp2p::{Multiaddr, PeerId};
 
 use crate::network::peers::handler::{Handler, HandlerFromBehaviour, HandlerToBehaviour};
-use crate::network::peers::Peer;
-use crate::replication::SyncMessage;
+use crate::network::peers::{Peer, PeerMessage};
 
 #[derive(Debug)]
 pub enum Event {
     /// Message received on the inbound stream.
-    MessageReceived(Peer, SyncMessage),
+    MessageReceived(Peer, PeerMessage),
 
     /// We established an inbound or outbound connection to a peer for the first time.
     PeerConnected(Peer),
@@ -105,7 +104,7 @@ impl Behaviour {
         &mut self,
         peer_id: PeerId,
         connection_id: ConnectionId,
-        message: SyncMessage,
+        message: PeerMessage,
     ) {
         let peer = Peer::new(peer_id, connection_id);
         self.push_event(ToSwarm::GenerateEvent(Event::MessageReceived(
@@ -130,7 +129,7 @@ impl Behaviour {
         self.enabled = true
     }
 
-    pub fn send_message(&mut self, peer: Peer, message: SyncMessage) {
+    pub fn send_message(&mut self, peer: Peer, message: PeerMessage) {
         self.push_event(ToSwarm::NotifyHandler {
             peer_id: peer.id(),
             event: HandlerFromBehaviour::Message(message),
@@ -236,7 +235,7 @@ mod tests {
     use p2panda_rs::schema::SchemaId;
     use rstest::rstest;
 
-    use crate::network::Peer;
+    use crate::network::{Peer, PeerMessage};
     use crate::replication::{Message, SyncMessage, TargetSet};
     use crate::test_utils::helpers::random_target_set;
 
@@ -306,7 +305,10 @@ mod tests {
         // Send a message from to swarm_1 local peer from swarm_2 local peer.
         swarm_1.behaviour_mut().send_message(
             Peer::new(swarm_2_peer_id, ConnectionId::new_unchecked(1)),
-            SyncMessage::new(0, Message::SyncRequest(0.into(), TargetSet::new(&vec![]))),
+            PeerMessage::SyncMessage(SyncMessage::new(
+                0,
+                Message::SyncRequest(0.into(), TargetSet::new(&vec![])),
+            )),
         );
 
         // Await a swarm event on swarm_2.
@@ -370,13 +372,19 @@ mod tests {
         // Send a message from swarm_1 to swarm_2
         swarm_1.behaviour_mut().send_message(
             peer_2,
-            SyncMessage::new(0, Message::SyncRequest(0.into(), target_set_1.clone())),
+            PeerMessage::SyncMessage(SyncMessage::new(
+                0,
+                Message::SyncRequest(0.into(), target_set_1.clone()),
+            )),
         );
 
         // Send a message from swarm_2 to swarm_1
         swarm_2.behaviour_mut().send_message(
             peer_1,
-            SyncMessage::new(1, Message::SyncRequest(0.into(), target_set_2.clone())),
+            PeerMessage::SyncMessage(SyncMessage::new(
+                1,
+                Message::SyncRequest(0.into(), target_set_2.clone()),
+            )),
         );
 
         // And again add the next behaviour events which occur in either swarms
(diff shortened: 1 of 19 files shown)


aquadoggo/src/replication/message.rs
```diff
@@ -13,32 +13,37 @@ use serde::{Deserialize, Serialize};
 
 use crate::replication::{Mode, SessionId, TargetSet};
 
-pub const SYNC_REQUEST_TYPE: MessageType = 0;
-pub const ENTRY_TYPE: MessageType = 8;
-pub const SYNC_DONE_TYPE: MessageType = 9;
+pub const ANNOUNCE_TYPE: MessageType = 0;
+pub const SYNC_REQUEST_TYPE: MessageType = 1;
+pub const SYNC_DONE_TYPE: MessageType = 2;
+pub const ENTRY_TYPE: MessageType = 3;
```
adzialocha, line 19: I've changed the identifiers here, not sure if this is how we want to do it but in the specification we didn't set an identifier for "Announce" messages.

aquadoggo/src/schema/schema_provider.rs
```diff
@@ -114,9 +112,26 @@ impl SchemaProvider {
...
+                .map(|schema| schema.id().to_owned())
+                .collect(),
+        }
+    }
+
+    /// Returns true if a whitelist of supported schema ids was provided through user
+    /// configuration.
+    pub fn is_whitelist_active(&self) -> bool {
+        self.whitelisted_schema_ids.is_some()
     }
```
adzialocha, line 135: It was a little bit hard for me to understand the difference between "supported schema ids" in their different states (with or without config provided), I hope that this API makes it a bit clearer?

aquadoggo_cli/src/schemas.rs
```diff
@@ -15,28 +16,3 @@ pub fn read_schema_ids_from_file(file: &mut File) -> Result<Vec<SchemaId>> {
...
-
-        create_dir_all(tmp_path.parent().unwrap()).unwrap();
-        let mut file = File::create(&tmp_path).unwrap();
-        file.write_all("schemas = [\"plant_0020c11ab63099193a8c8516cc00f88bfc8cdd657b94a99fef2fce86aaaede84f87d\"]".as_bytes()).unwrap();
-
-        let result = read_schema_ids_from_file(&mut file);
-        println!("{result:?}");
-        assert!(result.is_ok());
-    }
-}
```
adzialocha, line 42: That test somehow didn't create a temporary file on my local machine (also I think the method always returns an empty array as the key is wrong in the written test data). I didn't dive deeper into troubleshooting it as this will change soon with the new configuration stack (where we probably will use serde to deserialize the `config.toml` file).

aquadoggo/src/replication/service.rs
```diff
@@ -264,17 +273,18 @@ impl ConnectionManager {
...
     /// Determine if we can attempt new replication sessions with the peers we currently know
     /// about.
     async fn update_sessions(&mut self) {
         // Determine the target set our node is interested in
         let target_set = self.target_set().await;
 
-        if let Err(err) = target_set.validate() {
-            warn!("Not initiating replication: {err}");
-            return;
-        }
```
adzialocha, line 276: I would assume that our own target set is always valid (especially when it was created through the `new` constructor) or are there any cases I overlooked?

aquadoggo/src/network/peers/message.rs
```diff
@@ -0,0 +1,15 @@
...
+
+/// p2panda protocol messages which can be sent over the wire.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum PeerMessage {
+    /// Announcement of peers about the schema ids they are interest in.
+    Announce(AnnouncementMessage),
+
+    /// Replication status and data exchange.
+    SyncMessage(SyncMessage),
+}
```
adzialocha, line 15: We might need a specification change here because without a message type identifier for `Announce` it will be hard for serde to distinct between these variants.

adzialocha:
I don't feel like adding more tests to this, we have e2e tests checking against successful replication and I've extended our replication service "lifetime" test to deal with announcement messages. Think that should be sufficient!

aquadoggo/src/network/peers/behaviour.rs
```diff
@@ -12,13 +12,12 @@ use libp2p::swarm::{
...
 use crate::network::peers::handler::{Handler, HandlerFromBehaviour, HandlerToBehaviour};
-use crate::network::peers::Peer;
-use crate::replication::SyncMessage;
+use crate::network::peers::{Peer, PeerMessage};
 
 #[derive(Debug)]
 pub enum Event {
     /// Message received on the inbound stream.
-    MessageReceived(Peer, SyncMessage),
+    MessageReceived(Peer, PeerMessage),
```
sandreae, line 20: I like the naming become more general here :+1:

aquadoggo/src/replication/announcement.rs
```diff
@@ -0,0 +1,189 @@
...
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("System time invalid, operation system time configured before UNIX epoch")
+        .as_secs()
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Announcement {
+    /// This contains a list of schema ids this peer is interested in.
+    pub target_set: TargetSet,
```
sandreae, line 23: I wonder if it would be better to use different naming here rather than "target_set" which is maybe quite a low-level term which comes from the replication protocol. Maybe `supported_schema_ids` as we use in the config file, or something else more self-explanatory.

aquadoggo/src/replication/ingest.rs
```diff
@@ -49,12 +49,16 @@ impl SyncIngest {
 
         let plain_operation = decode_operation(encoded_operation)?;
 
-        // If the node has been configured with supported_schema_ids, check that the sent
-        // operation follows one of our supported schema.
-        if let Some(supported_schema_ids) = self.schema_provider.supported_schema_ids() {
-            if supported_schema_ids.contains(plain_operation.schema_id()) {
-                return Err(IngestError::UnsupportedSchema);
-            }
+        // If the node has been configured with a whitelist of supported schema ids, check that the
```
sandreae, line 52: Should change "whitelist" to "allowlist" here and any other places in this PR too.

aquadoggo/src/schema/schema_provider.rs
sandreae: Yeh, this is better, thanks :+1:

aquadoggo/src/replication/announcement.rs
sandreae: Hmm, after reading through the rest of the PR i see that announce is of course closely coupled to the rest of the replication service and protocol code. So maybe this naming is fine, interested to hear what you think though @adzialocha . I'm not interested in holding up this PR in order to find _the best_ name :rofl:

sandreae (changes requested):
I had a couple of comments, only about naming. Except for that, all looking really good!

aquadoggo/src/replication/ingest.rs
adzialocha: Yes! I've changed that in the config PR 👍🏻

aquadoggo/src/replication/announcement.rs
adzialocha: Good point! A target set is a subset of the supported ids, it should be named differently

Issue #516, opened 2023-08-20
Better "Initialised schema provider" logging info into printed config


adzialocha:
The logs showing:

```
[2023-08-20T21:55:46Z DEBUG aquadoggo::schema::schema_provider] Initialised schema provider:
    - schema_definition_v1
    - schema_field_definition_v1
```

.. is confusing as it does not reflect the schemas I'm actually supporting if I'm using `config.toml`.

Let's fix that and move it into the printed config on node startup.

Issue #513, opened 2023-08-18
Generate configuration from env vars, config file and command line arguments


adzialocha:
The priority is as follows (latest has the highest):

1. Config File `config.toml`
2. Environment variables
3. Command line arguments

PR #519, opened 2023-08-22
Improved configuration API with TOML file

### LINKED ISSUE #511: Simplify configuration arguments

adzialocha:
Closes #511, #513, #516, #510

## 📋 Checklist


jadehopepunk:
I'm not going to police the language in anyone's code, but some teams I'm in have moved away from the whitelist/blacklist terminology to allowlist/denylist

aquadoggo_cli/config.toml
```diff
@@ -0,0 +1,166 @@
...
+# and materializing documents connected to it, supporting applications which
+# are dependent on this data.
+#
+# It is recommended to set this list to all schema ids your own application
+# should support. Defaults to an empty list.
+#
+# WARNING: When set empty, your node will support _all_ schema ids. This is
+# useful for experimentation and local development but _not_ recommended for
+# production settings.
+#
```
jadehopepunk, line 27: The above is reallyy clear, great stuff

aquadoggo_cli/config.toml
```diff
@@ -0,0 +1,166 @@
...
+quic_port = 2022
+
+# ~~~~~~~~
+# Identity
+# ~~~~~~~~
+
+# Path to persist your ed25519 private key file. The key is used to identify
+# you towards other nodes during network discovery and replication. This key is
+# _not_ used to create and sign data.
+#
```
jadehopepunk, line 102: That last sentence helps clarify the questions I had, thanks

aquadoggo_cli/config.toml
```diff
@@ -0,0 +1,166 @@
...
+# WARNING: This will potentially expose your IP address on the network. Do only
+# connect to trusted relays or make sure your IP address is hidden via a VPN or
+# proxy.
+#
+# WARNING: Using relays will potentially connect you to untrusted / unknown
+# nodes with which you will then exchange data with. If in doubt, use the list
+# of known node addresses instead and only connect to trusted nodes.
+#
+# relay_address = "192.0.2.16:2022"
```
jadehopepunk, line 157: Is it really just one relay_address? The section is called "relays", so I'm wondering if it should be an array.

aquadoggo_cli/config.toml
```diff
@@ -0,0 +1,166 @@
...
+
+# Set to true if our node should also function as a relay. Defaults to false.
+#
+# Other nodes can use relays to aid discovery and establishing connectivity.
+#
+# NOTE: Relays _need_ to be hosted in a way where they can be reached directly,
+# for example with a static IP address through an VPS.
+#
+relay = false
```
jadehopepunk, line 166: I wonder if this might be clearer if it was called `im_a_relay` or something, to differentiate being a relay from using a relay.

aquadoggo_cli/config.toml
adzialocha: It's been an array before but then I realized we don't support this yet (is that right @sandreae), will clarify it in the wording!

adzialocha:
> I'm not going to police the language in anyone's code, but some teams I'm in have moved away from the whitelist/blacklist terminology to allowlist/denylist

All really helpful feedback, thank you! Agree with that weird terminology.

aquadoggo_cli/config.toml
sandreae: Yeh, I only accounted for connecting to one relay in the recent networking work. It doesn't look too bad to refactor to account for multiple relays, but it's not something I've spent time testing. I'll spend a bit of time on it this evening and see how it goes. Having multiple relays is the desired behaviour for sure.

aquadoggo_cli/config.toml
```diff
@@ -0,0 +1,167 @@
...
+# ~~~~~
+
+# List of addresses of trusted and known nodes.
+#
+# NOTE: Make sure that nodes mentioned in this list can be connected directly
+# (for example when they are hosted with a static IP Address). If you need to
+# connect to nodes with changing, dynamic IP addresses or even with nodes
+# behind a firewall or NAT, you will need at least one Relay.
+#
+node_addresses = [
```
sandreae, line 130: oops : https://github.com/p2panda/aquadoggo/issues/521

aquadoggo_cli/config.toml
```diff
@@ -0,0 +1,167 @@
...
+# dependent on this data.
+#
+# It is recommended to set this list to all schema ids your own application
+# should support. Defaults to an empty list.
+#
+# WARNING: When set empty, your node will support _all_ schema ids. This is
+# useful for experimentation and local development but _not_ recommended for
+# production settings.
+#
+supported_schema_ids = [
```
sandreae, line 28: It would be good to be able to configure a node which doesn't support any schema, this way you could run a relay which aids in discovery & connectivity but _doesn't_ itself replicate any data. Ideally by setting these two configuration args as so:

```
supported_schema_ids = []
im_a_relay = true
```

Could this be achieved if we allow a wildcard value in the schema ids list `[*]`?

sandreae (commented):
This is looking really great to me, I had a suggestion around being able to configure your node to not support any schema ids. 

Am i correct understanding that the default for a node without any custom configurations is that it is "ephemeral" (no db or identity state is persisted)?

sandreae:
Thanks for your comments @jadehopepunk all really useful!

aquadoggo_cli/config.toml
sandreae: Or make it so that when no `supported_schema_ids` value is provided (it's commented out in the config) then we assume this to mean _any_ are supported.

aquadoggo_cli/config.toml
adzialocha: Oh oh

aquadoggo_cli/config.toml
adzialocha: I'll think about the wildcard thingy, that would be the most explicit.

We can't write `[*]` because this is not valid toml but `"*"` would work.

adzialocha:
Adding this https://github.com/libp2p/rust-libp2p/tree/master/misc/allow-block-list and having a config for it would be quite nice. This would allow for better control with whom you're connecting with, even when relays are involved.

aquadoggo_cli/config.toml
sandreae: > This is the current behaviour 👍

Ah, ok, I misread that from the code then, thought the default of an empty set was to support everything.

sandreae: As long as there are ways to account for the three situations:

- support no schema
- dynamically support all schema
- support a set of specific schema

Then I think that's all we need, doesn't have to be with the wildcard approach, just omitting the support schema config is totally fine.

adzialocha: > Then I think that's all we need, doesn't have to be with the wildcard approach, just omitting the support schema config is totally fine.

Yes, this we already support. The wildcard would be an extra to make it more explicit.

aquadoggo_cli/src/config.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub mdns: bool,
+    pub direct_node_addresses: Vec<SocketAddr>,
+    pub relay_address: Option<SocketAddr>,
+    pub relay_mode: bool,
+}
+
+impl Default for Configuration {
+    fn default() -> Self {
+        Self {
+            supported_schema_ids: UncheckedAllowList::Set(vec![]),
```
sandreae, line 145: Been playing around with the cli (it's super fun) and I don't think it's possible to set `Wildcard` mode for supported_schema_ids.

sandreae: This would be quite useful, I'm looking into it now, let me know if i'm wrong in my understanding :+1:

sandreae: We could change the default to be `Wildcard` then all configurations will be available from the cli (only)

aquadoggo_cli/config.toml
sandreae: implemented here: https://github.com/p2panda/aquadoggo/pull/524

PR #524, opened 2023-08-25
Networking: multiple relays, dial known nodes, allow&block-list

sandreae:
- [x] client nodes don't need to listen on tcp
- [x] handle connecting to multiple relay nodes (w/ config changes)
- [x] dial known direct nodes at startup
- [x] introduce allow-list of trusted peers
- [x] introduce block-list of un-trusted peers

## 📋 Checklist

- [x] ~Add tests that cover your changes~


aquadoggo/src/network/behaviour.rs
```diff
@@ -116,7 +124,7 @@ impl P2pandaBehaviour {
 
         // Create a rendezvous client behaviour with default configuration if a rendezvous server
         // address has been provided
-        let rendezvous_client = if network_config.relay_address.is_some() {
+        let rendezvous_client = if !network_config.relay_addresses.is_empty() {
```
adzialocha, line 127: I want to rename this field to `relay_nodes`, but let's do this later

adzialocha (approved):

PR #532, opened 2023-08-25
Prepare README for v0.5.0 release

adzialocha:
Closes: https://github.com/p2panda/aquadoggo/issues/526

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <h1 align="center">aquadoggo</h1>
 
 <div align="center">
-  <strong>Embeddable p2p network node</strong>
+  <strong>p2panda network node</strong>
 </div>
 
 <br />
@@ -43,18 +43,11 @@
 
 <br/>
 
-Configurable node implementation for the [`p2panda`] network running as a
-[`command line application`] or embedded via the [`library`] inside your Rust
-program.
+`aquadoggo` is a reference node implementation for [p2panda](https://p2panda.org). It is a intended as a tool for making the design and build of local-first, collaborative p2p applications as simple as possible, and hopefully even a little fun!
 
-> The core p2panda [`specification`] is fully functional but still under review so
-> please be prepared for breaking API changes until we reach v1.0. Currently no
-> p2panda implementation has received a security audit.
+`aquadoggo` can run both on your own device for local-first applications, or on a public server when acting as shared community infrastructure. Nodes like `aquadoggo` perform a number of tasks ranging from core p2panda data replication and validation, aiding the discovery and establishment of connections between edge peers, and exposing a developer friendly API used for building applications.
 
-[`command line application`]: /aquadoggo_cli
-[`library`]: /aquadoggo
-[`p2panda`]: https://p2panda.org/
-[`specification`]: https://p2panda.org/specification
+Read more about nodes in our [learn](https://p2panda.org/learn/networks) section.
 
 ## Features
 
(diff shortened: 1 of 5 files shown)


adzialocha:
This looks really amazing to me now! Thank you @sandreae for all of these good ideas for the README.

adzialocha:
This I stumbled upon, what do you think? It's right at the beginning but directly diverges to other projects, I think that can be a bit confusing (people want to get started with aquadoggo and not with fishy)? Could the section just move down, or at least somehow the links part?

sandreae:
Yeh, totally agree that it's weird to link out to other resources right away. I had the idea to add a new section near the end of the README titled `Resources` and here we can list fishy, send-to-node, tutorials etc... Then we don't need to link to them directly in the section you highlighted.

sandreae:
I also want to add a section showing some GraphQL query examples.

README.md
```diff
@@ -43,65 +43,189 @@
...
 - Discovers other nodes in local network and internet.
-- Replicates data with other nodes.
+- Establishes peer-to-peer connections via UDP holepunching or via relays.
+- Replicates data efficiently with other nodes.
+
+## Who is this for?
+
+`aquadoggo` might be interesting for anyone who wants to participate in a p2p network. This could be as a node maintainer, an application developer or simply someone wanting to learn more about p2p networking in a hands-on fashion.
+
+No programming experience is needed to deploy a node on your own machine, and you can even experiment with creating your own data schemas using, publishing and replicating data, and then querying it again using the GraphQL playground.
```
adzialocha, line 67: "using"

README.md
```diff
@@ -43,65 +43,297 @@
...
+If you are not working with Rust you can create FFI bindings from the `aquadoggo` crate into your preferred programming language. Dealing with FFI bindings can be a bit cumbersome and we do not have much prepared for you (yet), but check out our [Meli](https://github.com/p2panda/meli/) Android project as an example on how we dealt with FFI bindings for Dart / Flutter.
+
+## Resources
+
+- 🐬 Deploy your own `aquadoggo` following the [tutorial](https://p2panda.org/tutorials/aquadoggo)
+- 🛠️ Create your own schemas using [`fishy`](https://github.com/p2panda/fishy)
+- 🛼 Open the GraphQL playground in your browser, served under `http://localhost:2020/graphql`
+- 📖 Try the [mushroom app tutorial](https://p2panda.org/tutorials/mushroom-app)
+- 🔬 Manually publish JSON encoded data to a node [`send-to-node`](https://github.com/p2panda/send-to-node)
+- 🐼 [Learn more](https://p2panda.org/learn/) about how p2panda works
```
adzialocha, line 322: Even though it's true it's confusing to mention JSON here maybe, people might think that p2panda data in general is JSON encoded

README.md
```diff
@@ -43,65 +43,297 @@
...
+A collection query for "events" which includes ordering and filtering as well as selecting some meta fields. Here only events between the specified dates and with a title containing the string 'funtastic' will be returned, they will be arranged in ascending chronological order:
+```graphql
+{
+  events: all_events_0020aaabb3edecb2e8b491b0c0cb6d7d175e4db0e9da6003b93de354feb9c52891d0(
+    first: 20
+    orderBy: "happening_at"
+    orderDirection: ASC
+    filter: {
+      title: { contains: "funtastic" }
+      happening_at: { gte: 2023-08-31, lte: 2023-09-14 }
```
adzialocha, line 201: Think the values need to be an array

sandreae: Which values?

sandreae: The `happening_at` values need to be unix timestamps

sandreae:
Aaaaaand actually I recon we need to beef up our doc strings in `lib.rs`, they still contain something very similar to the old README.

sandreae:
> Oh, that's pretty cool, examples in the included README are also run during tests (although it fails right now).

I had do disable this feature for now cos there is required code setup which we can't do in the README :sweat_smile:

sandreae (approved):

PR #544, opened 2023-09-01
HTTP routes to serve files with correct content type and etag headers

### LINKED ISSUE #541: Select current view for blob using custom middleware
this would mean we don't need to symlink between blobs and their current view

adzialocha:
This PR switches from using symlinks and the `ServeDir` struct from `tower` to HTTP routes which resolve the blobs manually and respond with the correct content types.

Notable changes:

* Files are now directly stored into the `/blobs` folder, without sub-directories using their document view id
* HTTP requests to `/blobs/:documentId` respond with the latest blob
* HTTP requests to `/blobs/:documentId/:viewId` respond with the blob at that view id (this makes `/blobs/documents/...` obsolete
* Responses contain `Content-Type` header set by the blob's self-declared MIME type
* Responses contain `ETag` header
* Requests with a `If-None-Match` header might be responded with a 304 "Not Modified" in case the entity tags matched
* Added `update_blob` test method
* Changed `add_blob` test method to slice up blob data into "pieces" by given "max piece length" variable
* Renamed test client method to reflect that it is not anymore for GraphQL only
* Changed all `fs` methods to async `tokio::fs`

Closes #541 and #501

## 📋 Checklist


aquadoggo/src/materializer/tasks/blob.rs
```diff
@@ -75,38 +73,36 @@ pub async fn blob_task(context: Context, input: TaskInput) -> TaskResult<TaskInp
...
             None => return Err(TaskError::Critical("No base path configured".to_string())),
         };
-
-        let blob_dir = base_path.join(blob_document.id().as_str());
-
-        fs::create_dir_all(&blob_dir).map_err(|err| TaskError::Critical(err.to_string()))?;
-        let blob_view_path = blob_dir.join(blob_document.view_id().to_string());
+        // @TODO: Why is this required here? Shouldn't the folder already exist at this point? (if
+        // I remove this writing a file fails ..)
+        fs::create_dir_all(base_path).unwrap();
```
adzialocha, line 85: This `create_dir_all` is a little bit of a mystery for me. If I remove it, creating a file fails, even though the folder should have been already created when the whole node got initialised?

aquadoggo/src/test_utils/node.rs
```diff
@@ -161,7 +161,7 @@ pub async fn add_document(
         // We only want to issue dependency tasks.
         let dependency_tasks = tasks
             .iter()
-            .filter(|task| task.worker_name() == "depenedency");
+            .filter(|task| task.worker_name() == "dependency");
```
adzialocha, line 164: 😅

sandreae: Guess we never really wanted to issue dependency tasks either 😅

aquadoggo/src/materializer/tasks/blob.rs
adzialocha: I would like to explore this a bit more when we have the configuration interface (where `base_path` might change into something outside of `/tmp`)

adzialocha: Change this now to an async tokio version of this method .. that seems to do the trick?

aquadoggo/src/materializer/tasks/blob.rs
```diff
@@ -75,38 +71,33 @@ pub async fn blob_task(context: Context, input: TaskInput) -> TaskResult<TaskInp
...
+            ))
+        })?;
+
+        file.write_all(&blob_data).await.map_err(|err| {
+            TaskError::Critical(format!(
+                "Could not write blob file @ {}: {}",
+                blob_view_path.display(),
+                err
+            ))
+        })?;
```
adzialocha, line 100: I've changed this to use the `tokio::fs` methods, this assures that writing to files is done in a separate thread and doesn't block the main thread :+1:

sandreae (changes requested):
Looks great to me, very exciting to have the last few blob features coming in.

sandreae (approved):

PR #547, opened 2023-09-04
Build a byte buffer over paginated pieces when assembling blobs

adzialocha:
This removes the maximum limit of allowed blob pieces (for now, we might want to have it back with some sort of configuration interface). This touched upon the pagination logic used to iterate over all pieces, so I decided to just paginate through it in a classic manner, page by page.

On top this uses now the `BytesMut` struct to efficiently assemble the blob pieces. For the future we might want to do this instead https://github.com/p2panda/aquadoggo/issues/548 to avoid keeping very large blobs in memory.

Closes: #546 

## 📋 Checklist

### DIFF
--- a/aquadoggo/Cargo.toml
+++ b/aquadoggo/Cargo.toml
@@ -27,6 +27,7 @@ asynchronous-codec = { version = "0.6.2", features = ["cbor"] }
 axum = { version = "0.6.10", features = ["headers"] }
 bamboo-rs-core-ed25519-yasmf = "0.1.1"
 bs58 = "0.4.0"
+bytes = "1.4.0"
 deadqueue = { version = "0.2.3", default-features = false, features = [
     "unlimited",
 ] }
--- a/aquadoggo/src/db/stores/blob.rs
+++ b/aquadoggo/src/db/stores/blob.rs
@@ -2,6 +2,7 @@
 
 use std::num::NonZeroU64;
 
+use bytes::{BufMut, BytesMut};
 use p2panda_rs::document::traits::AsDocument;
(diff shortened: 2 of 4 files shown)


aquadoggo/Cargo.toml
```diff
@@ -27,6 +27,7 @@ asynchronous-codec = { version = "0.6.2", features = ["cbor"] }
 axum = { version = "0.6.10", features = ["headers"] }
 bamboo-rs-core-ed25519-yasmf = "0.1.1"
 bs58 = "0.4.0"
+bytes = "1.4.0"
```
adzialocha, line 30: We already have this dependency with `asynchronous_codec` but I'm still adding it explicitly here

aquadoggo/src/db/stores/blob.rs
```diff
@@ -10,14 +11,11 @@ use p2panda_rs::storage_provider::traits::DocumentStore;
...
-use crate::db::query::{Field, Filter, Order, Pagination, Select};
+use crate::db::query::{Filter, Order, Pagination, PaginationField, Select};
 use crate::db::stores::query::{Query, RelationList};
 use crate::db::SqlStore;
 
-/// The maximum allowed number of blob pieces per blob.
-/// @TODO: do we want this? If so, what value should it be and we should add this to
-/// p2panda-rs blob validation too.
-const MAX_BLOB_PIECES: u64 = 10000;
+const BLOB_QUERY_PAGE_SIZE: u64 = 10;
```
adzialocha, line 18: Think it's cool to paginate over blob pieces, it's an optimization but really nice for very large files 😎

sandreae (approved):

PR #551, opened 2023-09-06
Stream blob data in chunks to files to not occupy too much memory

adzialocha:
Establishes a stream from reading binary blob data from the database (via pagination) into writing the file on the file system. This allows for (parallely) handling very large blobs without occupying too much of memory.

Closes: https://github.com/p2panda/aquadoggo/issues/548

## 📋 Checklist


sandreae (approved):
Really good, looking forward to exploring optimal configurations for managing memory use with this in place.

PR #552, opened 2023-09-06
Integrate `Bytes` operation value

sandreae:
MIGRATED TO: https://github.com/p2panda/aquadoggo/pull/554

- store `Bytes` in the database in a new "data" column in the `operation_field_v1` table (not supported by Postgres....)
- when handling fields of the type `Bytes` we look at this column instead of "value"
- in the `GraphQL` bytes are represented by `HexBytesScalar`

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/db/models/query.rs
+++ b/aquadoggo/src/db/models/query.rs
@@ -45,6 +45,10 @@ pub struct QueryRow {
     #[sqlx(default)]
     pub value: Option<String>,
 
+    /// Data field used to store blob_piece_v1 bytes
+    #[sqlx(default)]
+    pub data: Option<Vec<u8>>,
+
     /// Type of the application field (string, boolean, float, int, relation etc.).
     #[sqlx(default)]
     pub field_type: String,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -278,8 +278,8 @@ version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "726535892e8eae7e70657b4c8ea93d26b8553afb1ce617caee529ef96d7dee6c"
 dependencies = [
- "proc-macro2 1.0.66",
- "quote 1.0.31",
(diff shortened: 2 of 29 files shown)


sandreae:
SQLite has `BLOB` type for binary data, Postgres has `Large Object` or `bytea`....

aquadoggo/src/db/models/query.rs
```diff
@@ -45,6 +45,10 @@ pub struct QueryRow {
     #[sqlx(default)]
     pub value: Option<String>,
 
+    /// Data field used to store blob_piece_v1 bytes
+    #[sqlx(default)]
+    pub data: Option<Vec<u8>>,
```
adzialocha, line 50: Interesting pattern! I wonder if we have other types there one day as well.

I'd name this column `value_bytes` to underline it's origin from an operation field and maybe keep the door open to have other field types as well.

adzialocha:
> SQLite has `BLOB` type for binary data, Postgres has `Large Object` or `bytea`....

Time to drop support? 😥

adzialocha:
Did you consider not storing blob pieces values at all in the database and retrieve the bytes from the file instead during replication?

adzialocha:
> Did you consider not storing blob pieces values at all in the database and retrieve the bytes from the file instead during replication?

https://doc.rust-lang.org/std/io/trait.Seek.html 😎

sandreae:
> Did you consider not storing blob pieces values at all in the database and retrieve the bytes from the file instead during replication?
> 
> https://doc.rust-lang.org/std/io/trait.Seek.html 😎 .. and interesting for writing: https://doc.rust-lang.org/std/io/struct.Cursor.html
> 
> I don't think it is too hard but also definitely not required for this development period, I've made a ticket here: #553

Yeh, this would be ideal. We still store the encoded operation in the db, so we wouldn't actually need to grab the bytes during replication either. Don't know if I want to tackle it right now though but it could be soon :+1: .

PR #554, opened 2023-09-07
Integrate `Bytes` value

sandreae:
_rebased from https://github.com/p2panda/aquadoggo/pull/552_

- store `Bytes` in the database as a hexadecimal string still in the `operation_field_v1` "value" column
  - we considered moving it to a new "bytes_value" field and storing as a `BLOB` but this would break compatibility with Postgres and in any case we are likely to stream the bytes directly to and from the filesystem in a later optimisation 
- in the `GraphQL` api bytes are represented by `HexBytesScalar`

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/graphql/queries/collection.rs
+++ b/aquadoggo/src/graphql/queries/collection.rs
@@ -629,7 +629,10 @@ mod tests {
             add_document(
                 &mut node,
                 schema.id(),
-                vec![("bool", false.into()), ("data", vec![4, 5, 6, 7][..].into())],
+                vec![
+                    ("bool", false.into()),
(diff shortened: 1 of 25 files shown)


aquadoggo/src/graphql/queries/collection.rs
```diff
@@ -385,6 +387,7 @@ mod tests {
                     ("artist", "David Bowie".into(), None),
                     ("title", "Speed Of Life".into(), None),
                     ("release_year", 1977.into(), None),
+                    ("audio", vec![8, 9, 10, 11][..].into(), None),
```
adzialocha, line 390: Haha cool!

PR #587, opened 2023-11-02
Use `write_all`  when writing blob data to file system

### LINKED ISSUE #586: Materialised blobs sometimes corrupted / missing bytes
Sometimes blobs are materialised with missing bytes. Noticed mostly when uploading larger files.

sandreae:
We should use `write_all` instead of `write` when writing blob data to the filesystem as the latter only represents a "single attempt" to write a buffer to a file, whereas the former "continuously calls [write] until there is no more data to be written"  or an error occurs. In using the later we were sometimes silently failing to write all data to the file system. 

Closes #586 

## 📋 Checklist


adzialocha:
Really good to know ✨✊🏻

PR #598, opened 2023-11-17
Node API to migrate lock files

### LINKED ISSUE #597: `migrate` API to allow bootstrapping schema data on launch of node
```rust
use aquadoggo::{Configuration, Node};
use p2panda_rs::identity::KeyPair;

let config = Configuration::default();
let key_pair = KeyPair::new();
let node = Node::start(key_pair, config).await;

let data = include_str!("schema.lock");
let lock_file: LockFile = toml::from_str(&data).expect("error parsing schema.lock file");
node.migrate(&lock_file).await;
```

Related discussion: https://github.com/p2panda/fishy/pull/11

adzialocha:
Introduces a `migrate` method on the `Node` struct to programmatically allow schema migrations and seeds.

```rust
let data = include_str!("schema.lock");
let lock_file: LockFile = toml::from_str(&data).expect("error parsing schema.lock file");
let node = Node::start(KeyPair::new(), Configuration::default()).await;
node.migrate(lock_file).await.expect("Migration failed");
```

This is maybe also a first step towards a general "low-level" API to interact with `aquadoggo`, for this we might want a `NodeBuilder` pattern instead to configure the node, while developers can use then the resulting `Node` itself to interact, but since this is a larger API change we should think this through a bit more.

Closes #597 

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Introduce a `migrate` API on the Node to allow publishing data programmatically [#598](https://github.com/p2panda/aquadoggo/pull/598)
+
 ## [0.6.0]
 
 ### Added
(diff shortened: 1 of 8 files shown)


sandreae (approved):
Looks great to me! I tested it out a bit and didn't come across any issues.

Issue #602, opened 2023-11-28
allow_schema_ids config setting not working


jmanm:
# Description 
I'm attempting to replicate a schema from one node to another and getting this log message:  "Replication failed: Incoming data could not be ingested: Schema is not supported".

## Steps to reproduce
1. Clone the repo
2. Edit ``config.toml`` to persist the database:
```toml
database_url = "sqlite:db/db.sqlite3"
```
3. Start the node:
```bash
cargo run -- -c ./config.toml --log-level info
```
4. Add a schema to the node. I used the mushroom-app-tutorial schema.
5. Add the resulting schema id(s) to ``config.toml``:
```toml
allow_schema_ids = [
    "schema_definition_v1",
    "schema_field_definition_v1",
    "blob_v1",
    "blob_piece_v1",
    "mushroom_<your document ID>",
    "mushroom_finding_<your document ID>"
]
```
6. Copy ``target/debug/aquadoggo`` and ``config.toml`` into a new folder.
7. In this folder, start a 2nd node:
```bash
./aquadoggo -c ./config.toml --log-level info
```

Node 1 will attempt to replicate the schema to node 2, but this error is logged:
``WARN  aquadoggo::replication::service] Replication failed: Incoming data could not be ingested: Schema is not supported``
and the schema is not replicated.

adzialocha:
Thank you for raising this issue! This indeed looks like a logical bug.

PR #603, opened 2023-11-28
fix allowed schema ID logic; add test

jmanm:
Maybe I'm misunderstanding the logic, but it seems like the condition in ``ingest.rs`` should be reversed. This addresses #602.

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix bug where known schemas are not replicated between nodes [#602](https://github.com/p2panda/aquadoggo/pull/603).
+- Fix bug where known schemas are not replicated between nodes [#603](https://github.com/p2panda/aquadoggo/pull/603).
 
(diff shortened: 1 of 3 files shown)


adzialocha (changes requested):
Thanks so much @jmanm for this fix and additional test! Well spotted.

I'm happy to merge this, just one small thing: For whatever reason I can't run our CI against this PR, but I just tested it locally and indeed the format checker is not happy. Can you just run `cargo fmt` and push this again? Otherwise our CI will cry latest when we merge into `main`. Thank you!

jmanm:
> Thanks so much @jmanm for this fix and additional test! Well spotted.
> 
> I'm happy to merge this, just one small thing: For whatever reason I can't run our CI against this PR, but I just tested it locally and indeed the format checker is not happy. Can you just run `cargo fmt` and push this again? Otherwise our CI will cry latest when we merge into `main`.
> 
> Another minor nitpick: The PR number in the `CHANGELOG.md` is wrong (needs to be #603 instead of #602), the link is correct.
> 
> Thank you!

Done. Thanks for the feedback!

adzialocha (approved):

PR #605, opened 2023-12-09
Let document subscription stream updates from materializer

cafca:
The subscription field for receiving updates about a specific document is updated to trigger sending a new update whenever the materializer finishes a new version of it.

This required changing up the message bus and service manager a bit to allow receiving updates about newly updated documents from within the `http` service. `Context` receives a new field containing the service sender, which is now constructed outside the service manager and passed in to it as a parameter.

Then, a new message type `DocumentUpdated` is added, which is emitted in the reducer task whenever documents are created or updated. This lets use then connect any running subscription stream to the service bus and listen for updates of their queried document.

## 📋 Checklist

### DIFF
--- a/aquadoggo/src/bus.rs
+++ b/aquadoggo/src/bus.rs
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use p2panda_rs::document::DocumentId;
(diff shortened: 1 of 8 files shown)


sandreae:
This is already looking good :pray: :pray: :pray: ! Is there any more functionality you want to include?

Issue #613, opened 2024-06-04
Accept string addresses in all configuration structs


sandreae:
PR #612 introduced made it possible to pass arbitrary string addresses into the aquadoggo cli via args or a config file. Under the hood these are then parsed into socket addresses using `to_socket_addrs()` (including ip lookup if required) and then converted to `Multiaddr` if the lookup or parsing fails for any address the cli process exits.

We still require that addresses used in our programmatic configuration structs presented as `Multiaddr`. This means the above flow is still required when starting a node from within another application. I suggest the following steps to only require string addresses here as well.

- change `Multiaddr` to `String` in all configuration structs
- if `to_socket_addrs()` fails then stop the node with a useful error
- if `to_socket_addrs()` returns no addresses (ip lookup was performed and no addresses found) then issue a debug warning but _don't_ stop the node

PR #621, opened 2024-06-14
Introduce `PeerAddress` struct for improved address resolution patterns

sandreae:
In all public aquadoggo configuration APIs peers can be addressed by IP address or domain name, in the case of the later, before making a connection attempt we need to do a DNS lookup to get the correct socket address. When the node is offline then we want to silently ignore failed lookups, just log the attempt and continue. Currently we only perform lookup when a node starts and we parse the configuration file, but later we want to introduce redials of relays and direct nodes, this will require us to perform the DNS lookup later (it may have failed on startup). 

This PR just introduces a new struct `PeerAddress` with two methods `socket` and `quic_multiaddr` which help us handle still unresolved peer addresses. When `socket` is called and successfully resolves the peer address string to a `SocketAddr` the result is cached and directly used on later calls in order to avoid unnecessary DNS lookups.

closes: #613 

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Re-materialize blobs which were only partially written to disc due to node crash [#618](https://github.com/p2panda/aquadoggo/pull/618)
 
+### Added
+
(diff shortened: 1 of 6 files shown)


adzialocha (approved):
Very nice solution ❤️‍🔥

PR #622, opened 2024-06-14
Dial all configured known relay and direct node addresses on schedule

sandreae:
When a node is first launched it attempts to connect to any configured relay or direct node addresses, the steps are as follows:

1) register at all relay/rendezvous nodes for peer discovery and connections
2) connect to and initiate a replication sessions with all relay/rendezvous nodes
3) connect to and initiate a replication sessions with all direct nodes

If the device where a node is running is offline when the app starts, all of these steps will fail and no further connection attempts are made. Similarly, if the node loses connectivity after it started, connections will be dropped for good. 

This PR implements a modest solution for points `2)` an `3)` above by adding a polling service to the `EventLoop` which attempts to (re)connect to all known peer addresses (relay or otherwise) every x seconds. A connection is only established if none already exist to the address in question. Replication sessions will be initiated with any peers that we successfully (re)connect to.

Finding a solution in the same situation for point `1)` above is a little more involved, these changes seemed like easy low hanging fruit which brings greatly improved UX to node users, especially in situations where nodes run for long periods and connections may drop.

## 📋 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Introduce `PeerAddress` struct to help resolve `String` to internal address types [#621](https://github.com/p2panda/aquadoggo/pull/621)
+- Re-dial all configured known peers on schedule [#622](https://github.com/p2panda/aquadoggo/pull/622)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


adzialocha (approved):
Looking good!

On mobile phones we have ways to find out about the connection status of a device, it could be nice to have manual methods to disconnect and connect on the Node API instead of frequent checks, that could be a more efficient approach. Pragmatically this PR is totally fine though and solves the issue.

PR #635, opened 2024-06-24
Support private net with pre-shared key

sandreae:
Implement private network secured via a pre-shared key using libp2p [pnet](https://github.com/libp2p/rust-libp2p/tree/master/transports/pnet]. When in "private net" mode a node will only accept connections with peers after verifying they know the same pre-shared key. `pnet` itself is a transport through which a connection is (in libp2p terms) "upgraded". 

It is only possible to run a private net when using TLC as the base transport (QUIC not supported) and so if a pre-shared key is provided the node is forced into using TLC regardless of other configurations.

- [x] introduce `pnet` network behaviour
- [x] accept pre-shared key in configuration file and via cli
- [x] force TLC when pre-shared key provided
- [x] log warnings if QUIC transport was requested  
- [ ] network tests: _i tried to write some simple network tests but it doesn't seem possible without introducing a new swarm just for testing, I believe "memory" multiaddr would need to be supported in our core `P2pandaBehaviour`. This would require some additional exploration and as we're using pnet in it's basic setup (which is tested on the libp2p side) it seems ok to me that there are no tests in this place_   

## 📋 Checklist

- [x] ~Add tests that cover your changes~ (see above)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Support private network secured by pre-shared key [#635](https://github.com/p2panda/aquadoggo/pull/635) 
+
 ### Changed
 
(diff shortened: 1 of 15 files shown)


adzialocha (approved):
🌟 🌟 🌟 🌟 🌟

